PR #434, opened 2020-05-08
[WIP] Map page routing

thomasgwatson:
Pushed up dev branch clone of mycelia-map to be the base of the feature-branch all the map stuff can be merged to.

Trying to chip away at the map page and expose my assumptions/ignorance/questions. This PR is just a short way to validate that map specific routes look like how we want them to. The MapFeed component is just a port-and-adjust of the Feed component, working on the assumption that the Feed store/connector and layout will be a good basis for the inputs that the Map view will need.

Something we can use as a basis for our chat today


thomasgwatson:
Current output state:
Screen Shot 2020-05-10 at 11 15 17

Will need to merge Tibet's location branches to the evo and node repos to start adding posts to the map

src/components/MapExplorer/MapExplorer.js
```diff
@@ -0,0 +1,129 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import { throttle, isEmpty, some } from 'lodash/fp'
+// import cx from 'classnames'
```
tibetsprague, line 4: remove comments before completing

src/routes/MapFeed/MapFeed.connector.js
```diff
@@ -0,0 +1,100 @@
+import { connect } from 'react-redux'
```
tibetsprague, line 1: i wonder how much we can reuse Feed.connector somehow instead of duplicating so much...

src/components/Map/Map.stories.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import Map from 'components/Map'
+import createAskOfferH3Layer from './layers/askOfferH3Layer'
+
+let offersAndAsks = [
+  { type: 'ask', summary: 'Need help Sunday', coordinates: [37.3615593, -122.0553238], hex: '8b283470d921fff' },
```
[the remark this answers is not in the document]
thomasgwatson: At the moment these are just sample layers from the demo map. I just pulled over the last commit from the map demo branch that had the bare minimum map.

src/routes/MapFeed/MapFeed.connector.js
thomasgwatson: Originally I thought MapFeed would be doing a whole lot more but since then I've reconsidered. Want to get closer to actually implementing the whole map view before ripping it out.

Issue #418, opened 2020-04-21
Add map view 


tibetsprague:
Add a map view within Hylo to display all resources, requests and offers that have a location and are visible to the current user.  You can view this map within a community, network or all your communities. It is a new tab listed next to the feed and projects

https://www.figma.com/file/lh2DVGMh7CRdUsxF3k6vGI/Map-Improvements?node-id=128%3A384

- [x] Clicking on a pin on the map will bring up a card with info about the post: https://github.com/Hylozoic/hylo-evo/issues/445
- [x] Posts visible in the current map view are also listed along the side of the map in cards like Airbnb: https://github.com/Hylozoic/hylo-evo/issues/446
- [x] The map updates as its viewport moves or scales: https://github.com/Hylozoic/hylo-node/issues/479
- [x] The map can be searched using full text search and filtered by topic tag
- [x] Mousing over a post in the map shows tooltip: https://github.com/Hylozoic/hylo-evo/issues/444
- [x] Posts are clustered so at zoomed out scales they appear as grouped icons with a post count for the cluster: https://github.com/Hylozoic/hylo-evo/issues/450

thomasgwatson:
https://www.figma.com/file/lh2DVGMh7CRdUsxF3k6vGI/Map-Improvements?node-id=8%3A1
https://www.hylo.com/all/p/30121

Linked some related commentary

thomasgwatson:
@brodeur I have more specific design needs after reading into what is required for a clustered icon layer.

These layers need an icon atlas and a mapping to those icons. The icon atlas from the example icon-cluster map I have previously shared is below:

![image](https://user-images.githubusercontent.com/5193218/81092545-98159f00-8eb5-11ea-8993-87e07f4baeeb.png)

The layer then needs the mapping to be able to pick out the right icon to display. A subset of the mappings for the above icon

   {
     "marker-1": {
       "x": 0,
         "y": 0,
         "width": 128,
         "height": 128,
         "anchorY": 128
     },
     "marker-2": {
       "x": 128,
         "y": 0,
         "width": 128,
         "height": 128,
         "anchorY": 128
     },
   }

So, for one thing: we need every possible permutation of the icons packed into the library (unless we build an icon providing endpoint). We have to use png; css and svg belong to the DOM and these mapping libraries rely on webGL accelerated canvas elements.

[...]

Issue #413, opened 2020-04-18
Add locations to posts


tibetsprague:
For Requests and Offers and the upcoming Resources post type add an optional Location field. 

- [x] Don't show for discussions?
- [x] Should we use a location autocomplete field from Google?

Depends on https://github.com/Hylozoic/hylo-node/issues/463

tibetsprague:
using Geocoder component made for Mapbox

PR #438, opened 2020-05-13
Mycelia map

tibetsprague:


tibetsprague:
includes fixes for https://github.com/Hylozoic/hylo-evo/issues/413 and https://github.com/Hylozoic/hylo-evo/issues/418

thomasgwatson:
Catching up on this. 

Looks like bbox queries are ready on evo but don't see references to them on hylo-node? I was assuming they would need to be linked up on the backend; converting the new element of the graphQL query to a knex/postgres query addition. Just curious whether my assumption is true or whether there is already some auto-magic graphQL conversion

tibetsprague:
It’s here On the search-by-bounding-box Branch, just forgot to create the PR: https://github.com/Hylozoic/hylo-node/pull/484

Sent by m

src/components/Map/layers/clusterLayer/index.js
```diff
@@ -0,0 +1,126 @@
...
+  'cluster': { 'x': 0, 'y': 0, 'width': 30, 'height': 30 },
+  'event': { 'x': 30, 'y': 0, 'width': 20, 'height': 20 },
+  'member': { 'x': 50, 'y': 0, 'width': 20, 'height': 20 },
+  'offer': { 'x': 70, 'y': 0, 'width': 20, 'height': 20 },
+  'request': { 'x': 90, 'y': 0, 'width': 20, 'height': 20 },
+  'resource': { 'x': 110, 'y': 0, 'width': 20, 'height': 20 }
+}
+
+export function createIconLayerFromPostsAndMembers ({ boundingBox, members, posts, onHover, onClick }) {
+  // TODO: shouldn't need to filter by locationObject, should all have one for map...?
```
thomasgwatson, line 17: should all have one for map? Are all posts required to have locations?

src/components/Map/layers/clusterLayer/index.js
```diff
@@ -0,0 +1,126 @@
...
+  let data = posts.filter(post => post.locationObject && post.locationObject.center)
+    .map(post => {
+      return {
+        id: post.id,
+        type: post.type,
+        color: hexToRgb(FEATURE_TYPES[post.type].primaryColor),
```
thomasgwatson, line 23: This property isn't used yet, correct?

src/routes/MapExplorer/MapExplorer.connector.js
```diff
@@ -0,0 +1,129 @@
...
+    ], routeParams),
+    ...pick([
+      // TODO: these are actually not being queried by on the server, remove for now?
+      'sortBy',
+      'topic'
+    ], props),
+    boundingBox: state.MapExplorer.fetchPostsParam ? state.MapExplorer.fetchPostsParam.boundingBox : null
+  }
+
+  // TODO: maybe filtering should happen on the presentedPosts? since we do some of that presentation in the filtering code, like calling topics.toModelArray in the filters for every post each time
```
lorenjohnson, line 75: I'm always very up for a discussion on selectors and presenters in context of our ReduxORM world. I generally have the tactic of trying to get out of ReduxORM land as quickly as possible at the trunk (so within a selector) vs carrying-over any dependence on it to the component or connector level. Either way, I see nothing objectionable here in particular, and am just wondering what the benefit would be here of doing what you're considering?

src/routes/MapExplorer/MapExplorer.connector.test.js
```diff
@@ -0,0 +1,55 @@
...
+        hasMore: true,
+        pending: undefined,
+        posts: [
+          expect.objectContaining({ id: '1' }),
+          expect.objectContaining({ id: '3' }),
+          expect.objectContaining({ id: '2' })
+        ]
+      })
+    )
+  })
```
lorenjohnson, line 45: This looks just fine, and am curious if you would like to move away from snapshot testing for these cases or more generally? (We don't need to discuss here, just noting these conversations for later as I notice them)

src/routes/MapExplorer/MapExplorer.holochain.connector.js
```diff
@@ -0,0 +1,48 @@
+import { connect } from 'react-redux'
```
lorenjohnson, line 1: You can get rid of this file and update the `index.js`. I think it'll bomb anyway once you merge in dev if you merged the remove Holochain branch yet.

src/routes/MapExplorer/MapExplorer.js
```diff
@@ -0,0 +1,229 @@
+import React from 'react'
```
lorenjohnson, line 1: This all LGTM! And wondering what this component would be like if implemented with hooks? Like would it be more or less legible?

src/routes/MapExplorer/MapExplorer.scss
```diff
@@ -0,0 +1,179 @@
+.MapExplorer-container {
```
lorenjohnson, line 1: These class names are going to be already CSS Modules hash and name scoped to `MapExplorer` so you could scope names with that assumption.

src/routes/MapExplorer/MapExplorer.store.test.js
```diff
@@ -0,0 +1,20 @@
+import { fetchPosts } from './MapExplorer.store'
```
lorenjohnson, line 1: Ah, I see you embraced snapshots here so disregard my earlier question, this answers it enough.

src/routes/MapExplorer/index.js
```diff
@@ -0,0 +1,8 @@
+import { HOLOCHAIN_ACTIVE } from 'util/holochain'
```
lorenjohnson, line 1: ...as mentioned you can remove the Holochain stuff here.

src/routes/MapExplorer/MapExplorer.store.js
```diff
@@ -0,0 +1,395 @@
...
+export const FEATURE_TYPES = {
+  ...POST_TYPES,
+  member: {
+    primaryColor: '#2A4059', // $color-member
+    backgroundColor: '#FAFBFC', // $color-athens-gray
+    map: true
+  }
+}
+
+const communityPostsQuery = `query (
```
lorenjohnson, line 29: It could be nice to move these graphql queries out of this local store and into the `src/graphql`. I could go either way, but in terms of moving the project over time to a more idiomatic leveraging of graphql I started moving queries out these local "pod" `.store.js` files... It's again a longer discussion, if we ever want to dip into it... and this is all just fine!

src/store/reducers/queryResults.js
```diff
@@ -203,7 +203,8 @@ export function makeGetQueryResults (actionType) {
     // they are passed directly to a component. Should buildKey handle both
     // cases?
     const key = buildKey(actionType, props)
-    return get(`queryResults[${key}]`, state)
+    // NOTE: cannot use lodash.get here because boundingBox string includes [, ] and . characters which are special in get
+    return state.queryResults ? state.queryResults[key] : null
```
lorenjohnson, line 207: Good to know.

src/routes/MapExplorer/MapExplorer.scss
tibetsprague: yeah not sure how that happened

src/routes/MapExplorer/MapExplorer.js
tibetsprague: as soon as components get complicated i find it hard to stick with the pure function component

src/routes/MapExplorer/MapExplorer.holochain.connector.js
tibetsprague: done

src/routes/MapExplorer/MapExplorer.connector.test.js
tibetsprague: no i think we should stick with them!

src/routes/MapExplorer/MapExplorer.store.js
tibetsprague: interesting, seems like at that point we may just want to go full apollo... :)

PR #451, opened 2020-05-21
public posts

### LINKED ISSUE #416: Add option for a Post to be marked Public
To determine whether a post should be public or only visible in communities shared to.

- Only logged in users can create posts. 
- New posts must be posted in at least one community before they can also be selected as public.
- Not doing this for public post v0: There is a "Public" community that these posts are added to.
- This option is available for All post types

- [x] Add field to Post Editor to select Public in the communities dropdown
- [x] Style indication on posts of public vs private @brodeur 

Depends on https://github.com/Hylozoic/hylo-node/issues/466

cpolitano:
[closes #416 ]

- new line on Post Editor allowing user to toggle post as public or not
- Bonus: includes CommunitiesSelector dropdown enhancements such as better UX and Member Counts for each community


public/globe-icon.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
tibetsprague, line 1: oh yeah lets work on using the icon font

src/components/CommunitiesSelector/CommunitiesSelector.js
```diff
@@ -33,19 +33,25 @@ export default class CommunitiesSelector extends Component {
 
   findSuggestions = (searchText) => {
     const { options, selected } = this.props
-    const newSuggestions = differenceBy(options, selected, 'id')
-      .filter(o => o.name.match(new RegExp(searchText, 'i')))
-    this.setState({ suggestions: newSuggestions })
+    let newSuggestions
+    if (searchText && searchText > 0) {
```
tibetsprague, line 37: did you mean searchText.length? i could also see calling .trim on the searchText to ignore whitespace

src/components/KeyControlledList/KeyControlledList.js
```diff
@@ -106,7 +107,7 @@ export default class KeyControlledList extends React.Component {
   render () {
     const { selectedIndex } = this.state
 
-    const { theme, children, ...props } = this.props
+    const { tagType, theme, children, ...props } = this.props
```
tibetsprague, line 110: it would be nice if we could make this work in a more generic way since this is meant to be a generic component, so instead of tagType have props for headerText, and for countText, then check if the item has a .count field to know whether to show a count or not. but if that all is a pain then lets just document that we would like to change that in the future. like the FIXME comment that already exists here

src/components/PostEditor/PostEditor.js
```diff
@@ -425,16 +433,28 @@ export default class PostEditor extends React.Component {
...
+            <CommunitiesSelector
+              options={communityOptions}
+              selected={communities}
+              onChange={this.setSelectedCommunities}
+              readOnly={loading}
+              ref={this.communitiesSelector}
+            />
+          </div>
+        </div>
+        <PublicToggle togglePublic={this.togglePublic} isPublic={post.isPublic ? post.isPublic : false} />
```
tibetsprague, line 457: one sexy way to do `post.isPublic ? post.isPublic : false` is !!isPublic

src/components/PublicToggle/PublicToggle.js
```diff
@@ -0,0 +1,31 @@
...
+    isPublic: false
+  }
+
+  render () {
+    const { isPublic, togglePublic } = this.props
+
+    return (
+      <div styleName={cx('publicToggleContainer', { postIsPublic: isPublic })}>
+        <div styleName='publicToggle'>
+          <SwitchStyled checked={isPublic} onChange={togglePublic} backgroundColor={isPublic ? '#0DC39F' : '#8B96A4'} />
```
tibetsprague, line 24: nice, using the new component Tom made :)

src/components/TagInput/TagInput.js
```diff
@@ -103,21 +103,24 @@ export default class TagInput extends Component {
...
-    const { suggestions, className, theme, readOnly, maxTags, addLeadingHashtag, renderSuggestion } = this.props
+    const { suggestions, className, theme, readOnly, maxTags, addLeadingHashtag, renderSuggestion, tagType } = this.props
     if (!tags) tags = []
     if (!placeholder) placeholder = 'Type...'
 
     const optionalHashtag = addLeadingHashtag ? '#' : ''
 
     const selectedItems = uniqBy('id', tags).map(t =>
-      <li key={t.id} className={theme.selectedTag}>
+      <li key={t.id} className={theme.selectedTag} styleName='list-item-selected'>
```
tibetsprague, line 123: confusing to use both className and styleName, why is that needed? we already have a class on it called selectedTag. we could use the classNames library if we need two classes on it, one default and one from the theme. the different is that styleName looks up the className in whatever css file has been imported, className you have to specific the variable for the imported styles, otherwise they are the same

src/components/TagInput/TagInput.js
```diff
@@ -133,19 +136,24 @@ export default class TagInput extends Component {
         : [{ name: `no more than ${maxTags} allowed`, isError: true }]
       : suggestions
 
-    return <div className={cx(theme.root, { [theme.readOnly]: readOnly }, className)} onClick={this.focus}>
-      <ul className={theme.selected}>
+    return <div className={cx(theme.root, { [theme.readOnly]: readOnly }, className)} styleName='list-outer-container' onClick={this.focus}>
```
tibetsprague, line 139: same here and below

src/components/TagInput/TagInput.js
```diff
@@ -133,19 +136,24 @@ export default class TagInput extends Component {
...
+        <div className={theme.searchInput} styleName='list-inner-container'>
           <input
             className={theme.searchInput}
-            styleName={cx({ 'error': maxReached })}
+            styleName={cx({ 'error': maxReached }, 'list-input')}
             ref={this.input}
             type='text'
             placeholder={placeholder}
             spellCheck={false}
+            onClick={() => { this.handleChange('') }}
```
tibetsprague, line 152: should this be onFocus? What they if tab into the field? actually im curious what this is needed for too

src/store/presenters/presentPost.js
```diff
@@ -3,13 +3,23 @@ export default function presentPost (post, communityId) {
...
+
   return {
     ...post.ref,
     creator: post.creator,
     linkPreview: post.linkPreview,
     location: post.location,
+    isPublic: post.isPublic,
     commenters: post.commenters.toModelArray(),
     communities: post.communities.toModelArray(),
+    communitiesArray: communities,
```
tibetsprague, line 22: i think this could get confusing, to have both communities and communitiesArray. not sure the best solution. would it be hard to create a Community model object and add it to the modelArray instead?

cpolitano: Solution - make PostCommunities into connected component and move functionality from PresentPost to there.

src/components/CommunitiesSelector/CommunitiesSelector.js
cpolitano: yes, sounds good 👍

src/components/KeyControlledList/KeyControlledList.js
cpolitano: If it's ok with you, I'll just write a FIXME note for now, and I'll address it when I get to improvements on Topics/Tags

src/components/PostEditor/PostEditor.js
cpolitano: 💯

src/components/PublicToggle/PublicToggle.js
cpolitano: yeah it works great!

src/components/TagInput/TagInput.js
cpolitano: agreed, using classNames library is better 👍

src/components/TagInput/TagInput.js
cpolitano: 👍

src/components/TagInput/TagInput.js
cpolitano: yup onFocus is better, I just wasn't thinking about tabbing

src/components/KeyControlledList/KeyControlledList.js
tibetsprague: sure

public/globe-icon.svg
cpolitano: done

PR #454, opened 2020-05-23
Map tweaks

thomasgwatson:
A little bit of follow up on map related stuff


thomasgwatson:
Just a few tweaks to the map related stuff.

PR #458, opened 2020-05-25
Switch styled

thomasgwatson:
Merging straight into dev, since these changes only impact Storybook.

Just wanted to knock this component out in some spare time. The original switch has a different look and feel and doesn't have color customization; as it is already in use, I decided to build out a new switch and leave the old in place.

Based on https://www.figma.com/file/lh2DVGMh7CRdUsxF3k6vGI/Map-Improvements?node-id=0%3A1

Screen Shot 2020-05-25 at 14 23 04
Screen Shot 2020-05-25 at 14 22 59

### DIFF
--- a/src/components/Switch/Switch.stories.js
+++ b/src/components/Switch/Switch.stories.js
@@ -1,0 +1,19 @@
+import React, { useState } from 'react'
+import { storiesOf } from '@storybook/react'
+import { MemoryRouter } from 'react-router'
+import Switch from 'components/Switch'
+
+storiesOf('Switch', module)
+  .addDecorator(story => (
+    <MemoryRouter>{story()}</MemoryRouter>
+  ))
+  .add('Basic', () => <SwitchDemo />)
+
+function SwitchDemo () {
+  const [switchBool, setSwitchFool] = useState(true)
+
+  function handleToggleSwitch () {
+    setSwitchFool(!switchBool)
+  }
+  return (<Switch value={switchBool} onClick={handleToggleSwitch} />)
+}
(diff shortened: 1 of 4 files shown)


src/components/Switch/Switch.stories.js
```diff
@@ -0,0 +1,31 @@
...
+    setSwitchBool(!switchBool)
+  }
+  return (<Switch value={switchBool} onClick={handleToggleSwitch} />)
+}
+
+function SwitchStyledDemo () {
+  const [switchBool, setSwitchBool] = useState(false)
+
+  function handleToggleSwitch (checked, name) {
+    console.log(checked, name)
```
tibetsprague, line 27: remove console

src/components/SwitchStyled/SwitchStyled.js
```diff
@@ -0,0 +1,26 @@
...
+      <span styleName={cx('button', { buttonChecked: checked })} />
+    </div>
+  )
+}
+
+SwitchStyled.defaultProps = {
+  checked: false,
+  disabled: false,
+  backgroundColor: '#ff44ff',
+  onChange: (checked, name) => { console.log(checked, name) }
```
tibetsprague, line 23: remove console

thomasgwatson: This is a default function, to help developers as they are setting up their toggles. The alternative is a noop: () => {}. Would you prefer that?

src/components/Switch/Switch.stories.js
thomasgwatson: This is in the storybook, it is logging just to demonstrate how the component works

tibetsprague: ok, fair enough

src/components/SwitchStyled/SwitchStyled.js
tibetsprague: hmm, not sure, I guess you're right, this could be useful. go ahead with it

PR #463, opened 2020-05-27
Even better scrolling layout

tibetsprague:


tibetsprague:
fixed by responsive web work

PR #481, opened 2020-06-04
Strip rich text formatting on paste into DraftJS

lorenjohnson:


lorenjohnson:
> so i think this good for now, but i believe we are going to want to add some ability to do formatting soon. do you know the reason why rich text was breaking things?

Yes, it was breaking because there was a remaining `<u></u>` after the link preview was removed in Google Calendar pasted text. '<u>' tags are not whitelisted in our sanitize function (see ref below). I didn't go further into understanding why it switched from a `<a></a>` tag to that upon removing the link preview as the more basic issue was that we were allowing richtext in paste while not accounting for it at all downstream.

We are currently sanitizing out most any formatting on saving to the backend, so currently allowing richtext here provides no benefit that I'm aware of. Draftjs has a `handlePastedText` prop for a function to handle the sanitization and filtering of formatting in pasted text once we're ready to visit that. At that time we'd probably also want to re-visit this https://github.com/Hylozoic/hylo-utils/blob/58a851d881a9929efebfc43e9cbda8816ba553f9/src/text.js#L18 in concert with that as well as the detail templates.

tibetsprague:
gotcha, thanks!

lorenjohnson:
> gotcha, thanks!

I should also note for the future that Draftjs can do a lot more for us with little effort, there was a lot of restraint exercised in the current post detail creation more by design than capacity, if that makes sense. Hope that is all helpful :)

PR #512, opened 2020-06-16
First pass on styling everything

brodeur:


public/icon/filter.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
tibetsprague, line 1: in the latest I have added this to the icon font and to the drawer as an <Icon> component

public/icon/menu.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="14" height="11" viewBox="0 0 14 11" fill="none" xmlns="http://www.w3.org/2000/svg">
```
tibetsprague, line 1: I can show you how to add these too

src/routes/MapExplorer/MapDrawer/MapDrawer.js
```diff
@@ -41,6 +41,7 @@ function MapDrawer (props) {
       querystringParams={querystringParams}
       post={post}
       styleName='contentCard'
+      constrained={'true'}
```
tibetsprague, line 44: you want this to just be {true} not {'true'} though it will work with the string because that evaluates to true

src/routes/MapExplorer/MapExplorer.js
```diff
@@ -139,7 +138,8 @@ export default class MapExplorer extends React.Component {
         zoom={zoom}
       />
       <button styleName={cx('toggleDrawerButton', { 'drawerOpen': showDrawer })} onClick={this.toggleDrawer}>
-        <Icon name='Stack' green={showDrawer} styleName='icon' />
+        <img styleName={cx('openDrawer')} src='/icon/menu.svg' />
```
tibetsprague, line 141: dont need to use cx if its just one class, its just styleName='openDrawer', but also we should replace these with <Icon>s

src/routes/MapExplorer/MapDrawer/MapDrawer.js
brodeur: Linter gave me ann error withh {true}, after googling I put in {'true'} and linter stopped complaining

public/icon/menu.svg
brodeur: ok

PR #513, opened 2020-06-16
Update tests

lorenjohnson:


lorenjohnson:
@tibetsprague ok, this is ready to go. Not a very careful preening of the snapshots, but there wasn't much to tend to there. I think we're good!

tibetsprague:
oh I did notice that the LocationInput.test.js has a MapBox token in it, which is not ideal. though of course it could easily be snagged by someone using the app, so also not too big a deal... hmm, thoughts?

lorenjohnson:
> oh I did notice that the LocationInput.test.js has a MapBox token in it, which is not ideal. though of course it could easily be snagged by someone using the app, so also not too big a deal... hmm, thoughts?

As I didn't have this token in my test env it of course actually is getting removed in this PR. However, most generally it would probably be best to either mock the import in the test or, probably what I would do, add  `mapbox` as a component prop with a default value assigned to the imported library. Then in the test we can just provide mocked values for that prop and remove the vulnerability of having the value make it's way back into the snapshots... If you want to make that change it would be better, but I'm ok with it as is.

tibetsprague:
oh wait never mind, they were removed :)

PR #523, opened 2020-06-23
Event invitee options scoped by selected communities

lorenjohnson:
Requires: https://github.com/Hylozoic/hylo-node/pull/500

In addition to things mention in attached tickets this update includes addition of debounce for invitee search in both edit and add contexts, better visibility of "Invite" link in Post Card view, some refactoring, and somewhat more testing. 

IMPORTANT: Until we're done QA'ing this I've left the feature flag in place. So to release this to production we need to set the following ENV variable: `FEATURE_FLAG_EVENTS=on`

\* NOTE: The notifications story related to inviting has not been carefully QA'd.

### DIFF
--- a/src/components/PostEditor/PostEditor.js
+++ b/src/components/PostEditor/PostEditor.js
@@ -429,6 +429,9 @@ export default class PostEditor extends React.Component {
             <MemberSelector
               initialMembers={members || []}
               onChange={this.updateProjectMembers}
+              // NOTE: Turn-on to limit Project Member selection
+              // to the membership of the selected communities
+              // forCommunities={communities}
               readOnly={loading}
             />
           </div>
@@ -479,8 +482,8 @@ export default class PostEditor extends React.Component {
             <MemberSelector
               initialMembers={eventInvitations || []}
               onChange={this.updateEventInvitations}
-              readOnly={loading}
               forCommunities={communities}
+              readOnly={loading}
             />
           </div>
         </div>}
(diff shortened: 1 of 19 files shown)


src/components/PostEditor/PostEditor.js
```diff
@@ -429,6 +429,9 @@ export default class PostEditor extends React.Component {
             <MemberSelector
               initialMembers={members || []}
               onChange={this.updateProjectMembers}
+              // NOTE: Turn-on to limit Project Member selection
```
tibetsprague, line 432: i say lets turn this on. i dont actually understand how event/project invites currently work with people outside the communities it is posted to. Can they see the post even though they are not in any of the communities it is posted to?

lorenjohnson: Yes, I think this was always a known gap and due to the lift required--which I did this week--we had let it slip and were planning to loop back. Now that this is done it makes sense to catch it up here. * However, the tidy part of my brain wants to push this branch/PR and then apply this to projects as a separate branch/PR (I can get that out pretty quickly on Monday).

lorenjohnson:
We good to merge? * Once it's released to master we need to set the `FEATURE_FLAG_EVENTS=on` on hylo-evo production, but we don't want to set that until this release is on production so as to not expose the currently buggy version, if that makes sense. 

* I'm out camping and out of range from now'ish til I'm back online on Sunday night or Monday AM (my time). I'm ok with this going out now, but if anything comes-up you won't be able to find me until then probably.

tibetsprague:
@lorenjohnson why wouldnt we just remove the need for the FEATURE_FLAG completely? i guess i wanted to turn on for projects too before we merge but otherwise yes!

tibetsprague:
ok, decided to just merge this as is, and keep testing. and yes do a separate branch for projects

PR #527, opened 2020-06-26
Mycelia responsive

brodeur:


src/routes/AllTopics/AllTopics.js
```diff
@@ -83,7 +83,7 @@ export default class AllTopics extends Component {
 
     const { totalTopicsCached } = this.state
 
-    return <FullPageModal>
+    return <FullPageModal topicView>
```
tibetsprague, line 86: i wish there was a way to avoid having page specific props like this. could we change the name to represent more what it is changing about the style for this page? in this case could be something like `fullWidth`?

src/routes/Messages/Messages.js
```diff
@@ -24,9 +25,17 @@ export default class Messages extends React.Component {
   constructor (props) {
     super(props)
 
+    let messagesOpen = false
+    if (props.messageThreadId && !props.isInbox) {
+      messagesOpen = true
+    } else if (props.isInbox) {
```
tibetsprague, line 31: you dont need this condition since it is already set to false above

src/routes/Messages/ThreadList/ThreadList.js
```diff
@@ -25,15 +26,21 @@ export default class ThreadList extends Component {
...
       className
     } = this.props
 
     return <div styleName='thread-list' className={className}>
       <div styleName='header'>
-        <Link to='/t/new'><Button label='New Message' styleName='new-message' /></Link>
+        <div styleName='closeMessages'>
+          <CloseMessages onCloseURL={onCloseURL} />
+        </div>
+        <Link to='/t/new' onClick={toggleMessages}><Button label='New Message' styleName='new-message' /></Link>
```
tibetsprague, line 43: this seems weird to me, a Button inside a Link that has both a url destination and an onClick handler. lets look at this together, i want to understand it

src/routes/Messages/ThreadList/ThreadList.js
```diff
@@ -97,7 +106,7 @@ export function ThreadListItem ({
   const { names, avatarUrls } = participantAttributes(thread, currentUser, 2)
 
   return <li styleName='list-item'>
-    <Link to={`/t/${id}`}>
+    <Link to={`/t/${id}`} onClick={toggleMessages}>
```
tibetsprague, line 109: this too

src/routes/NonAuthLayout/NonAuthLayout.js
```diff
@@ -49,18 +49,21 @@ export default class NonAuthLayout extends React.Component {
       } />}
 
       {!HOLOCHAIN_ACTIVE && <Route path='/login' component={() =>
-        <Login {...this.props} styleName='form' />
+        <div styleName='signupRow'>
```
tibetsprague, line 52: i think you want this div inside the <Route> because you dont want it to display unless its on the <Login> route, right?

src/routes/NonAuthLayout/NonAuthLayout.js
```diff
@@ -49,18 +49,21 @@ export default class NonAuthLayout extends React.Component {
...
+          <Login {...this.props} styleName='form' />
+        </div>
       } />}
-
-      <Route path='/signup' component={() =>
-        <Signup {...this.props} styleName='form' />
-      } />
+      <div styleName='signupRow'>
```
tibetsprague, line 56: same

src/routes/PrimaryLayout/PrimaryLayout.js
```diff
@@ -170,7 +169,7 @@ export default class PrimaryLayout extends Component {
       </Switch>
       <SocketListener location={location} />
       <SocketSubscriber type='community' id={get('slug', community)} />
-      <Intercom appID={isTest ? null : config.intercom.appId} hide_default_launcher={hasDetail} />
+      <Intercom appID={isTest ? null : config.intercom.appId} hide_default_launcher />
```
tibetsprague, line 172: this is just always hiding it right? so why not just remove?

src/components/DownloadAppModal/DownloadAppModal.scss
```diff
@@ -10,6 +10,7 @@
   justify-content: center;
   z-index: 200000;
   margin: auto;
+  display: none;
```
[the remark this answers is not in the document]
brodeur: yes, how to best modify it... needs a ticket cuz it has to be rebuilt to do what we want, which is more like reddit. Prompt people to download but allow them to continue

src/components/FeedBanner/FeedBanner.scss
```diff
@@ -131,3 +136,41 @@
...
+    height: 50px;
+  }
+
+  .prompt-image {
+    margin-left: 10px;
+    top: 11px;
+  }
+}
+
+@media screen and (max-width: 375px) {
```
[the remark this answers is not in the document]
brodeur: done

src/routes/AllTopics/AllTopics.js
brodeur: Got it

src/routes/Messages/Messages.js
brodeur: k

src/routes/NonAuthLayout/NonAuthLayout.js
brodeur: done

src/routes/PrimaryLayout/PrimaryLayout.js
brodeur: This is where it is invoked, as far as I can tell. I can move it somewhhere else?

PR #528, opened 2020-06-27
Fix jpeg images uploading rotated in weird ways

tibetsprague:
Required updating to latest filestack APIs, which is good


lorenjohnson:
Great! 👍  LGTM, except the broken test

Issue #483, opened 2020-06-05
Can't edit Resource type posts


tibetsprague:
The save button stays disabled even as you edit the post

PR #529, opened 2020-06-27
Edit resource type posts

wushenshen:
Closes https://github.com/Hylozoic/hylo-evo/issues/483

### DIFF
--- a/src/components/PostEditor/PostEditor.js
+++ b/src/components/PostEditor/PostEditor.js
@@ -93,6 +93,7 @@ export default class PostEditor extends React.Component {
     })
     const currentPost = post
       ? ({ ...post,
+        locationId: post.locationObject.id,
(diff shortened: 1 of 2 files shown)


wushenshen:
btw @tibetsprague you mentioned possibly removing the location requirement for Resource type posts - that would involve removing [this line](https://github.com/Hylozoic/hylo-evo/blob/438de95f79071204e53275f1cacc46517f544c2a/src/components/PostEditor/PostEditor.js#L323) and perhaps changing [this line](https://github.com/Hylozoic/hylo-evo/blob/438de95f79071204e53275f1cacc46517f544c2a/src/components/PostEditor/PostEditor.js#L96) to `(post.locationObject || {})` so as to prevent errors when locationObject is not populated.

PR #545, opened 2020-07-01
469 non logged in view

tibetsprague:
Can view /public/map while not logged in to Hylo
Also can filter map by one or more networks using /public/map?network=xxx&network=yyy


src/routes/MapExplorer/MapExplorer.js
```diff
@@ -192,14 +191,14 @@ export default class MapExplorer extends React.Component {
 
   render () {
     const {
+      currentUser,
       features,
       fetchPostsParam,
       filters,
       querystringParams,
       pending,
       routeParams,
```
lorenjohnson, line 200: Passing `queryStringParams` and `routeParams` smells a little funny to me. If the component being rendered isn't part of the routing, and therefore doesn't have a `match.params` prop then for clarity's sake I might expect to pass only the specific keys (networkSlug, slug/communitySlug, etc) instead of the whole thing, starting from the connector in this component... That said, these leaf components are mostly route endpoints (and to me that is a good thing), and therefore should have access to `match.params`.  It is an option in those cases to retrieve these params directly within the component (and without connecting the component to Redux--which can be confusing as they are setup as selectors). It's just something I noticed. It can be confusing but worth keeping in mind, as some additional simplicity and clarity can be achieved.

src/routes/NonAuthLayout/Login/Login.connector.js
```diff
@@ -6,9 +7,10 @@ import { getReturnToURL, resetReturnToURL } from 'router/AuthRoute/AuthRoute.sto
 import mobileRedirect from 'util/mobileRedirect'
 
 export function mapStateToProps (state, props) {
+  const queryParams = qs.parse(props.location.search.substring(1))
```
lorenjohnson, line 10: I believe `getQuerystringParam` would be appropriate and simplifying here too, i.e.:

`const returnToUrl = getQuerystringParam('returnToUrl') || getReturnToUrl(state)`

src/routes/PostDetail/Comments/CommentForm/CommentForm.js
```diff
@@ -51,33 +55,40 @@ export default class CommentForm extends Component {
...
-    if (!currentUser) return null
-
-    const placeholder = `Hi ${currentUser.firstName()}, what's on your mind?`
+    const placeholder = currentUser ? `Hi ${currentUser.firstName()}, what's on your mind?` : "Hi! What's on your mind?"
 
     return <div
       styleName='commentForm'
       className={className}
       onClick={() => this.editor.current.focus()}>
       <AttachmentManager type='comment' id='new' />
```
lorenjohnson, line 64: Probably shouldn't render this either in the case of `!currentUser` though I think it's largely immaterial.

tibetsprague: actually it looks nice to have the comment box there but unusable and with a button that says Sign up to comment

lorenjohnson: Oh for sure, in total agreement. I mean the AttachmentManager is not necessary to render (there is nothing visible about it when there are no attachments)

lorenjohnson:
### QA Report

**Bugs**

- [ ] Upvoting Post doesn't do anything but acts like it does
- [ ] Clicking on "Public" in the communities list on the Post detail leads to infinite load page
- [ ] `/public` goes to infinite load page. Forward /public to /public/map, for now?

**Notes / nice to haves:**

- [ ] Map sidebar in public view allows filtering of results by topic, but the actual Post Detail sends user to Login if they click on a topic. Would be nifty if clicking on a topic in Post Detail forwarded to a search/filter of the map results for the currently selected area filtered by that tag. 
- [ ] Bookmark'able URL which centers on a locaiton and zoom level
- [ ] "Flagging" a post does actually succeed through to API. This isn't wrong, just noting.
- [ ] It would  be good for commenters on a Public post to have highlighted that their comments will appear publicly.

src/routes/PostDetail/Comments/CommentForm/CommentForm.js
tibetsprague: oh right

src/routes/MapExplorer/MapExplorer.js
tibetsprague: hmm yeah i hear you. the routeParams is used by PostCard and passed in all over the place to it so dont feel like doing that refactor right now, but i think i agree

tibetsprague:
@lorenjohnson clicking on Public in the community details brings me to the public stream which is working for me

tibetsprague:
which is the same as /public, which again, works for me

tibetsprague:
ohhh, you mean when not logged in. right

tibetsprague:
hmm, yeah i want to get the non-logged in stream view working but maybe too much work right now, so yeah redirecting to map seems like a good idea

tibetsprague:
flagging posts is supposed to work, we terrans get the message :)

tibetsprague:
https://github.com/Hylozoic/hylo-evo/issues/606

tibetsprague:
https://github.com/Hylozoic/hylo-evo/issues/607

tibetsprague:
https://github.com/Hylozoic/hylo-evo/issues/593

Issue #471, opened 2020-06-02
Allow for resources without locations


tibetsprague:
For online resources like articles, videos, books...

PR #574, opened 2020-07-24
Remove location requirement for resource type posts

wushenshen:
#471


wushenshen:
@tibetsprague ah yes nice catch, I updated

Issue #566, opened 2020-07-20
Add map search box to jump to a location on the map


tibetsprague:

thomasgwatson:
@tibetsprague Is this search in the figma design or will I have to use some creative license to find a position for it?

tibetsprague:
There are designs in Figma though there may be some questions left to figure out with Aaron. Also definitely feel free to take initiative on pieces not designed and then we can discuss

PR #580, opened 2020-07-26
Add search box to MapExplorer to allow location navigation

thomasgwatson:
Initial pass at this

#566


thomasgwatson:
Expecting that I will have to munge the LocationInput component in a way that doesn't save the searched location in the database (since that isn't useful in this instance).

thomasgwatson:
Preview:

Screen Shot 2020-07-25 at 18 14 28
Screen Shot 2020-07-25 at 18 14 40

PR #603, opened 2020-08-13
Fixes attachments in post edit, ordering of image attachments

lorenjohnson:
Fixes the issue with images attachments not appearing in Post edit, and issues in both Post and Comment editors with retaining image ordering.


tibetsprague:
18:28:29 web.1      |      at Function.createForSubject (/Users/tibet/Dev/hylo/hylo-node/api/models/Media.js:73:29)

tibetsprague:
same with trying to add an image to a post

lorenjohnson:
The editing post images is now fixed in the attached PR and the updating issue was a bug in the refactoring on the backend branch which is now fixed and on the `dev` branch of `hylo-node`.

PR #654, opened 2020-10-09
623 Re-arrange account and profile settings, 622 Add contactEmail, contactPhone to currentUser queries

lorenjohnson:


src/routes/CommunitySettings/CommunitySettingsTab/CommunitySettingsTab.scss
```diff
@@ -95,7 +95,8 @@
 
 label {
   color: $color-regent-70;
-  margin-left: 10px;
+  // LEJ: Commenting-out as has global impact
```
tibetsprague, line 98: why not just change 'label' to a classname?

src/util/navigation.js
```diff
@@ -180,6 +184,15 @@ export function newMessageUrl () {
   return `${messagesUrl()}/new`
 }
 
+export function messageThreadUrl (person) {
+  // TODO: messageThreadId doesn't seem to be currently ever coming-in from the backend
```
tibetsprague, line 188: what does this do on the front-end? file a ticket?

src/routes/CommunitySettings/CommunitySettingsTab/CommunitySettingsTab.scss
lorenjohnson: Yes, certainly. That does require making updating to this feature area and doing a more comprehensive review of the impact of removing the global style throughout, but yes I can and will do that.

src/util/navigation.js
lorenjohnson: https://github.com/Hylozoic/hylo-evo/issues/655

src/routes/CommunitySettings/CommunitySettingsTab/CommunitySettingsTab.scss
tibetsprague: hmm yeah, fair enough, would rather get this out right now

tibetsprague (approved):

src/routes/CommunitySettings/CommunitySettingsTab/CommunitySettingsTab.scss
lorenjohnson: NOTE: If this is going out as is, previous to the profile design update (in progress), Skill & Interests won't currently be appearing anywhere on the Profile.

I'm ok with that, the design update won't be far behind, but worth pointing-out.

---
... and yeah, :) basically that was the logic I was following as well. I did make better a few things in the course of this branch as it was but needed to stop there.

lorenjohnson:
@tibetsprague with your permission I'd like to add skill tags under bio as in this screenshot to this branch to make it more releasable as-is. This is only a few lines diff and a snapshot update on the PR:

Screen Shot 2020-10-09 at 8 54 32 AM

tibetsprague:
Love it! Go go go.

lorenjohnson:
> Love it! Go go go.

Ok it's in there now. Here is the remaining diff / final commit if you want to give it a quick look. Otherwise, now I think this would be a happy little on-the-way update. https://github.com/Hylozoic/hylo-evo/pull/654/commits/160dcd289dd54c7aac8ebca4e36e1afa94e688dc

tibetsprague (approved):

PR #656, opened 2020-10-09
621 new profile layout

lorenjohnson:
Requires https://github.com/Hylozoic/hylo-node/pull/540 and https://github.com/Hylozoic/hylo-node/pull/542


src/components/Pillbox/Pillbox.scss
```diff
@@ -42,7 +34,7 @@ $search-width: 266px;
     outline: none;
     color: rgba(44, 64, 89, 0.8);
     font-size: 15px;
-    border-bottom: 1px solid rgba(44, 64, 89, 0.3);
+    // border-bottom: 1px solid rgba(44, 64, 89, 0.3);
```
tibetsprague, line 37: remove comment

src/css/global/icons.scss
```diff
@@ -13,180 +13,192 @@
   -moz-osx-font-smoothing: grayscale;
 }
 
-.icon-EmailNotification:before {
+.icon-Letter:before {
```
tibetsprague, line 16: hmm, looks like you changed all the icons, in the future its nicer to add new icons to the end so we can see exactly what was added/changed. also it means i will have to redo my own icon collection with the new order

src/routes/MemberProfile/MemberProfile.js
```diff
@@ -33,60 +38,106 @@ export default class MemberProfile extends React.Component {
...
-      </ProfileBanner>
+      <div styleName='header'>
+        <div styleName='header-banner' style={bgImageStyle(person.bannerUrl)}>
+          <RoundImage styleName='header-member-avatar' url={person.avatarUrl} xlarge />
+          <h1 styleName='header-member-name'>{person.name}</h1>
+          {person.location && <div styleName='header-member-location'>
+            <Icon name='Location' styleName='header-member-location-icon' />
+            {locationWithoutUsa}
+          </div>}
+          {/* TODO: Do we still want to show the "Community manager" role? */}
```
tibetsprague, line 88: oh dang, forgot to work this out with aaron. i think it's a good thing to show but we dont have a design for it yet

src/routes/MemberProfile/MemberProfile.js
```diff
@@ -33,60 +38,106 @@ export default class MemberProfile extends React.Component {
...
+          />
+        </div>
+        <CurrentContentComponent routeParams={routeParams} loading={contentLoading} />
       </div>
     </div>
   }
 }
 
+export function handleContactPhone (contactPhone) {
+  return window.location.assign(`mailto:${contactPhone}`)
```
tibetsprague, line 134: should be tel:

src/routes/PrimaryLayout/PrimaryLayout.js
```diff
@@ -25,7 +25,7 @@ import Feed from 'routes/Feed'
 import MapExplorer from 'routes/MapExplorer'
 import Loading from 'components/Loading'
 import MemberProfile from 'routes/MemberProfile'
-import MemberSidebar from 'routes/MemberSidebar'
+// import MemberSidebar from 'routes/MemberSidebar'
```
tibetsprague, line 28: delete?

src/routes/PrimaryLayout/PrimaryLayout.js
```diff
@@ -161,11 +161,11 @@ export default class PrimaryLayout extends Component {
         <div styleName={cx('sidebar', { hidden: (hasDetail || isMapViewPath(location.pathname)) })}>
           <Switch>
             <Route path={`/:context(c)/:slug${OPTIONAL_NEW_POST_MATCH}`} exact component={CommunitySidebar} />
-            <Route path={`/:context(c)/:slug/m/:personId/${OPTIONAL_NEW_POST_MATCH}`} component={MemberSidebar} />
+            {/* <Route path={`/:context(c)/:slug/m/:personId/${OPTIONAL_NEW_POST_MATCH}`} component={MemberSidebar} /> */}
```
tibetsprague, line 164: we are just getting rid of this no?

src/css/global/icons.scss
lorenjohnson: That shouldn't be the case, I think you may be having a caching thing. Tell me more.

src/routes/PrimaryLayout/PrimaryLayout.js
lorenjohnson: I left this in in case you guys didn't like not having a sidebar. The layout seems a little strange in the "All Communities" context. Take a look, but either way though I'll remove it.

src/routes/MemberProfile/MemberProfile.js
lorenjohnson: Oh yeah, good catch. Copy/paste error.

src/routes/MemberProfile/MemberProfile.js
lorenjohnson: Ok

src/routes/PrimaryLayout/PrimaryLayout.js
tibetsprague: yeah i agree it looks a little weird. hmm, not sure what it should be

src/css/global/icons.scss
tibetsprague: yeah thats true, clearing the cache does fix it, though that could be an issue on deployment for people. thats why i would prefer we dont change all the icon codes, only add new ones to the end

lorenjohnson: I need to add another icon for the Copy tooltip thing and this time re-did the others so that the original codes were used for existing icons. It would be nice if we didn't have to be this careful in an icon updating process that is already a bit tacky, but for now I don't want to risk testing my contention that the caching will be busted in the webpack > production pipeline... SO we can both not worry about it now :)

src/routes/MemberProfile/MemberProfile.js
lorenjohnson: This isn't going to make this iteration of the feature. There seemed to be some confusion about it's meaning anyway, so maybe we can wait for it to come back into discussion and create a ticket for it then.

lorenjohnson:
This feature is done save the following:

* [ ] Facebook links are really long as they are API proxied links to the user's profile, not simply the "facebook.com/lorenjohnson" link which otherwise I'd expect. I will probably "..." shorten these URLs but haven't done that yet.

* [ ] Layout of profile in "All Communities" view. I see you created another ticket for this but may try and fix it here on my next and final pass.

tibetsprague (approved):
looks good to me! only thing i noticed is hovering over the facebook link you cant click on the link itself to go there. but im ok with that if its a pain

PR #662, opened 2020-10-14
Signing-up users stuck on blank screen

lorenjohnson:


tibetsprague (approved):
huh, just about route order?

lorenjohnson:
> huh, just about route order?

Yep. The signup routes were below the sign-up redirect. This probably wouldn't interfere if you completed the sign-up on the first pass without ever leaving or reloading.

PR #687, opened 2020-10-29
660 signup avatar upload improvement

lorenjohnson:


lorenjohnson:
@tibetsprague this makes the whole profile Avatar bubble clickable for upload on the Sign-up flow vs just the icon (as it is on in Edit User Profile. I think its a better better UX. Merge if you agree, close if not 👍

tibetsprague (approved):

PR #693, opened 2020-11-02
Add projects section to member profile

### LINKED ISSUE #631: Show which Projects a person is a part of on their profile
This means projects that the person has "joined" which will include ones they created.

Only show the Projects that the viewer of the profile is allowed to see

https://www.figma.com/file/zAjzEqnfPjEU3zUUdBOrBI/Profile?node-id=1%3A221

wushenshen:
Closes #631

### DIFF
--- a/src/components/RoundImageRow/RoundImageRow.js
+++ b/src/components/RoundImageRow/RoundImageRow.js
@@ -5,7 +5,7 @@ import './RoundImageRow.scss'
 
 const { array, string, bool } = PropTypes
 
-export default function RoundImageRow ({ imageUrls = [], onProfile = false, className, vertical, cap, ascending, ...rest }) {
+export default function RoundImageRow ({ imageUrls = [], inline = false, className, vertical, cap, ascending, ...rest }) {
   var capped
   var extra
   if (cap && cap < imageUrls.length) {
@@ -26,7 +26,7 @@ export default function RoundImageRow ({ imageUrls = [], onProfile = false, clas
       styleName='image'
       style={zIndexStyle(i)} />)
 
-  const plus = <div styleName={onProfile ? 'plus-profile' : 'plus-default'} key='plus' style={zIndexStyle(imageUrls.length)} >
+  const plus = <div styleName={!inline ? 'plus' : 'plus-inline'} key='plus' style={zIndexStyle(imageUrls.length)} >
     +{extra}
   </div>
 
--- a/src/components/RoundImageRow/RoundImageRow.js
+++ b/src/components/RoundImageRow/RoundImageRow.js
@@ -5,7 +5,7 @@ import './RoundImageRow.scss'
 
 const { array, string, bool } = PropTypes
 
-export default function RoundImageRow ({ imageUrls = [], className, vertical, cap, ascending, ...rest }) {
+export default function RoundImageRow ({ imageUrls = [], onProfile = false, className, vertical, cap, ascending, ...rest }) {
   var capped
   var extra
   if (cap && cap < imageUrls.length) {
@@ -26,7 +26,7 @@ export default function RoundImageRow ({ imageUrls = [], className, vertical, ca
       styleName='image'
       style={zIndexStyle(i)} />)
 
-  const plus = <div styleName='plus' key='plus' style={zIndexStyle(imageUrls.length)} >
+  const plus = <div styleName={onProfile ? 'plus-profile' : 'plus-default'} key='plus' style={zIndexStyle(imageUrls.length)} >
     +{extra}
   </div>
 
--- a/src/components/RoundImageRow/RoundImageRow.scss
+++ b/src/components/RoundImageRow/RoundImageRow.scss
@@ -1,26 +1,22 @@
-.plus-default {
+@mixin plus {
   position: relative;
   width: $space-9x;
   height: $space-9x;
   line-height: $space-7x;
   border-radius: 50%;
   border: 3px solid $color-white;
-  composes: caption-lt-lg from 'css/typography.scss';
   text-align: center;
+}
+
+.plus { 
+  @include plus();
   margin-top: -$space-4x;
   background: $color-caribbean-green;
   color: $color-white;
 }
 
-.plus-profile {
-  position: relative;
-  width: $space-9x;
-  height: $space-9x;
-  line-height: $space-7x;
-  border-radius: 50%;
-  border: 3px solid $color-white;
-  composes: caption-lt-lg from 'css/typography.scss';
-  text-align: center;
+.plus-inline {
+  @include plus();
   left: -15px;
   font-size: 0.8rem;
   margin-top: 0;
--- a/src/components/RoundImageRow/RoundImageRow.scss
+++ b/src/components/RoundImageRow/RoundImageRow.scss
@@ -1,15 +1,31 @@
-.plus {
+.plus-default {
   position: relative;
   width: $space-9x;
   height: $space-9x;
   line-height: $space-7x;
-  margin-top: -$space-4x;
   border-radius: 50%;
   border: 3px solid $color-white;
   composes: caption-lt-lg from 'css/typography.scss';
+  text-align: center;
+  margin-top: -$space-4x;
   background: $color-caribbean-green;
   color: $color-white;
+}
+
+.plus-profile {
+  position: relative;
+  width: $space-9x;
+  height: $space-9x;
+  line-height: $space-7x;
+  border-radius: 50%;
+  border: 3px solid $color-white;
+  composes: caption-lt-lg from 'css/typography.scss';
   text-align: center;
+  left: -15px;
+  font-size: 0.8rem;
+  margin-top: 0;
+  background: $color-athens-gray-dark;
+  color: $color-regent;
 }
 
 .image {
--- a/src/graphql/queries/personQuery.js
+++ b/src/graphql/queries/personQuery.js
@@ -44,6 +44,21 @@ export default
         slug
       }
     }
+    projects {
+      items {
+        id
+        title
+        createdAt
+        creator {
+          name
+        }
+        members {
+          items {
+            avatarUrl
+          }
+        }
+      }
+    }
     skills (first: 100) {
       total
       hasMore
--- a/src/routes/MemberProfile/MemberProfile.js
+++ b/src/routes/MemberProfile/MemberProfile.js
@@ -62,7 +62,7 @@ export default class MemberProfile extends React.Component {
       showDetails,
       push
     } = this.props
-    const { projects } = person
+    const projects = person.projects && person.projects.items
     const { currentTab } = this.state
     const personId = routeParams.personId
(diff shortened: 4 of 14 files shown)


src/components/RoundImageRow/RoundImageRow.scss
```diff
@@ -1,15 +1,31 @@
...
   border: 3px solid $color-white;
   composes: caption-lt-lg from 'css/typography.scss';
+  text-align: center;
+  margin-top: -$space-4x;
   background: $color-caribbean-green;
   color: $color-white;
+}
+
+.plus-profile {
+  position: relative;
```
tibetsprague, line 16: looks like a lot of duplicate stuff from the one above, should put the common stuff in a shared class. also, i wonder if the additional/changed styles are better handled by passing in a classname to the component

src/components/RoundImageRow/RoundImageRow.js
```diff
@@ -26,7 +26,7 @@ export default function RoundImageRow ({ imageUrls = [], className, vertical, ca
       styleName='image'
       style={zIndexStyle(i)} />)
 
-  const plus = <div styleName='plus' key='plus' style={zIndexStyle(imageUrls.length)} >
+  const plus = <div styleName={onProfile ? 'plus-profile' : 'plus-default'} key='plus' style={zIndexStyle(imageUrls.length)} >
```
tibetsprague, line 29: instead of having onProfile its more flexible to just allow for passing in a className/styleName to this component and then the styles can be specified in the parent component

src/routes/MemberProfile/MemberProfile.js
```diff
@@ -37,7 +39,9 @@ export default class MemberProfile extends React.Component {
 
   componentDidMount () {
     const { personId } = this.props.routeParams
+    const { currentUser } = this.props
     if (personId) this.props.fetchPerson(personId)
+    this.props.fetchProjects('MEMBER_PROFILE', personId, currentUser.id)
```
tibetsprague, line 44: should be in conditional for if personId

tibetsprague (changes requested):

src/components/RoundImageRow/RoundImageRow.scss
wushenshen: I tried doing this:

```
%plus {
  position: relative;
  width: $space-9x;
  height: $space-9x;
  line-height: $space-7x;
  border-radius: 50%;
  border: 3px solid $color-white;
  composes: caption-lt-lg from 'css/typography.scss';
  text-align: center;
}


.plus-default {
  @extend %plus;
  margin-top: -$space-4x;
  background: $color-caribbean-green;
  color: $color-white;
}

.plus-profile {
  @extend %plus;
  left: -15px;
  font-size: 0.8rem;
  margin-top: 0;
  background: $color-athens-gray-dark;
  color: $color-regent;
}
```

and got the error message `composition is only allowed when selector is single :local class name not in "%plus"`

src/components/RoundImageRow/RoundImageRow.js
wushenshen: @tibetsprague so you are thinking of passing two classNames to this component, since one goes for the parent div and then another is used on the internal "plus" div?

src/components/RoundImageRow/RoundImageRow.scss
lorenjohnson: Look at how to use SCSS `@mixins` in `MemberProfile.scss`, that would be one way to handle that. 

Otherwise, better, you can do this:

```
.plus { 
  position: relative;
  width: $space-9x;
  height: $space-9x;
  line-height: $space-7x;
  border-radius: 50%;
  border: 3px solid $color-white;
  text-align: center;
  // default props overriden below
  margin-top: -$space-4x;
  background: $color-caribbean-green;
  color: $color-white;
  &-profile { // makes .plus-profile
    left: -15px;
    font-size: 0.8rem;
    margin-top: 0;
    background: $color-athens-gray-dark;
    color: $color-regent;
  } 
}
```
You can then apply `plus` or `plus plus-profile`, respectively (or whatever sent-in classname applied).

src/components/RoundImageRow/RoundImageRow.js
lorenjohnson: You could create a `template` or `styles` prop that was an hash with classnames under the `plus` and `container` keys (or similar). With something like that you can optionally use a named import of styles.scss as `styles` in the parent component and directly send the whole styles hash table to the component to work with (e.g. `<RoundImageRow ... styles={styles} />`)

You could also leave the className applying over the container element and combining with the optional template provided style... 

\* I've always wished far leaf components to be style-less as much as possible, but it's often not so easy to do so. I like the template pattern I'm suggesting, but other devs I've ran that past often find it confusing or don't feel the extra complication is warranted.

src/routes/MemberProfile/MemberProfile.js
```diff
@@ -218,6 +225,33 @@ export function ActionDropdown ({ items }) {
...
+      {projects.map((p, index) => {
+        return (<Project key={index} project={p} showDetails={showDetails} />)
+      })}
+    </div>
+  )
+}
+
+export function Project ({ project, showDetails }) {
+  const { title, id, createdAt, creator, members } = project
+  const MEMBER_CAP = 3
```
lorenjohnson, line 245: Trivial note, but I would make this a prop with a default value, just because I think that'd be a little more clear, testable, and suddenly configurable.

src/routes/MemberProfile/MemberProfile.scss
```diff
@@ -213,7 +213,74 @@
...
+  border-radius: 5px;
+  cursor: pointer;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  min-height: 60px;
+  padding: 8px 11px;
+  margin: 7px auto;
+
+    .title {
```
lorenjohnson, line 243: I'm low-key wishing the indentation was the standard (in Hylo stuff so far) 2 spaces here instead of 4. Because standards, nothing more.

src/routes/MemberProfile/MemberProfile.store.js
```diff
@@ -5,6 +5,52 @@ export const FETCH_RECENT_ACTIVITY = 'FETCH_RECENT_ACTIVITY'
 export const FETCH_MEMBER_POSTS = 'FETCH_MEMBER_POSTS'
 export const FETCH_MEMBER_COMMENTS = 'FETCH_MEMBER_COMMENTS'
 export const FETCH_MEMBER_VOTES = 'FETCH_MEMBER_VOTES'
+export const FETCH_PROJECTS = 'FETCH_PROJECTS'
+
+const defaultState = {
+  projects: []
+}
+
+export default function reducer (state = defaultState, action) {
```
lorenjohnson, line 14: This is breaking a pattern in `hylo-evo` in that the results aren't getting passed into a ReduxORM model and selected from there, which is what we're doing for everything else. Projects to date were handled as a Post with additional attributes in the ReduxORM modeling. I am all for treating it as a separate type but splitting from ReduxORM is something we should be aware of and discuss I think. 

\* Opinion: I would like to focus the use of component reducers to places where they're only absolutely necessary.

lorenjohnson (changes requested):
Despite the opinions and nit picks here, things look just fine with the exception of a question on the local reducer / lack of ReduxORM usage for getting to the projects data. I think that is worth a pause and discuss.

src/components/RoundImageRow/RoundImageRow.scss
wushenshen: Ok, I have updated `RoundImageRow.scss` to use mixins

src/components/RoundImageRow/RoundImageRow.js
wushenshen: I think this is extra complexity not needed for now since there are only two variants, one where the plus circle is on the next line, and one where the plus circle is inline (used on the profile page for now). I've changed the boolean to be `inline` rather than `onProfile` to reflect this.

tibetsprague: great

src/routes/MemberProfile/MemberProfile.store.js
tibetsprague: ok, after more though I agree that we should stick with the ReduxORM pattern currently in use. @lorenjohnson if you are down to pair with @wushenshen on it that would be much appreciated, though i can also do that

wushenshen: moving the projects to the Person type fixes this issue too

src/routes/MemberProfile/MemberProfile.js
```diff
@@ -218,6 +223,32 @@ export function ActionDropdown ({ items }) {
     />
 }
 
+export function ProjectsSection ({ memberCap, projects, showDetails }) {
+  return (
+    <div>
+      <div styleName='profile-subhead'>
```
lorenjohnson, line 229: I think this heading would be better in the parent markup as done with the SkillsSection? This heading so far seems to me seems business of the MemberProfile layout. This may eliminate the use for this `ProjectsSection` component entirely which I think isn't a bad thing...

src/routes/MemberProfile/MemberProfile.store.js
```diff
@@ -20,6 +20,7 @@ export function getRoleForCommunity (person, communitySlug) {
 export function presentPerson (person, selectedCommunitySlug) {
   return {
     ...person.ref,
+    projects: person.projects && person.projects.items,
```
lorenjohnson, line 23: I wouldn't flatten the `items` pagination setup here. I think it could trip someone up later when/if there is need to implement the pagination on projects in this feature or elsewhere.

lorenjohnson (changes requested):
Consider my couple of comments. I think those would be things worth adjusting. Otherwise looks good!

wushenshen:
@lorenjohnson updated again! @tibetsprague also need another look from you in response to the changes you requested

src/graphql/queries/personQuery.js
```diff
@@ -44,6 +44,21 @@ export default
         slug
       }
     }
+    projects {
```
tibetsprague, line 47: @lorenjohnson this is where I actually start to get concerned about performance, if every time we do a person query we load all this stuff. this is what I think is one of the things making Hylo slow as compared to loading data as needed when its needed. for example the User profile settings page is now very slow. so... we may someday want to go back to @wushenshen original implementation with separate queries, but for now this is fine :)

tibetsprague (approved):
yay!

wushenshen:
FYI from testing locally it appears that the current functionality is to display only the posts (including projects) that are visible to the logged in user.

src/graphql/queries/personQuery.js
lorenjohnson: The projects will only be queried when they are queried.

lorenjohnson (approved):
👍  LGTM!

PR #696, opened 2020-11-06
Updates to work with upgraded hylo-node and heroku stack-18

tibetsprague:


lorenjohnson (approved):
👍  LGTM. Will that date change eliminate that long console warning maybe? Anyway, this all looks great. Thanks for getting through this!

Issue #632, opened 2020-09-03
Choose skills that you want to learn on your profile


tibetsprague:
Indicate a different between skills you have and ones you want to learn

https://us02web.zoom.us/j/81928846076?pwd=c3UzRVgrdzdnMlNWRThJY0k0SnFUZz09

PR #701, opened 2020-11-20
632 skills to learn

tibetsprague:
Not my favorite solution but best i could figure out for now

Fixes https://github.com/Hylozoic/hylo-evo/issues/632


lorenjohnson:
(See https://github.com/Hylozoic/hylo-node/pull/562#issuecomment-735419379)

tibetsprague:
@lorenjohnson ready for review again

lorenjohnson (approved):
👍  Approving.

One note to address or consider: 

I'm wondering why not re-use `SkillsSection` vs copy / paste. 

This could be accomplished by moving the graphql into the root of `graphql/mutations/addSkill|addSkillToLearn` (etc), keeping `components/SkillsToLearn/SkillsToLearn.connector.js` which would attach those respective queries, and reusing everything else from `.../SkillsToLearn`. . . Or similar.

This is not a common pattern in our code, but something in the Holo fork we ended-up doing more often. The reuse here would be appropriate as the uniformity of the UX is on these two features what is desirable (at least for now).

tibetsprague:
i did start by trying to reuse code but there are enough places where it would require awkward passing around of which type of skill is this and conditionals that it ended up feeling just as clean to copy even though there is significant code reuse. i am thinking about it again...

tibetsprague:
ok code now re-used

PR #702, opened 2020-12-01
Add upcoming events to profile

### LINKED ISSUE #698: Show upcoming events they are planning to attend on a person's profile
https://www.figma.com/file/zAjzEqnfPjEU3zUUdBOrBI/Profile?node-id=1%3A221

wushenshen:
Closes #698

### DIFF
--- a/src/graphql/queries/personQuery.js
+++ b/src/graphql/queries/personQuery.js
@@ -33,6 +33,25 @@ export default
       neighborhood
       region
     }
+    eventsAttending {
+      items {
+        id
+        title
+        location
+        startTime
+        eventInvitations {
+          items {
+            response
+            person {
+              avatarUrl
+            }
+          }
+        }
+        communities {
+          slug
+        }
+      }
+    }
     messageThreadId
     memberships {
       id
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- Events section to member profile, showing upcoming events the person is attending
+
 ## [2.2.2] - 2020-12-09
 ### Added
 - Added Skills to Learn to user profiles
@@ -13,7 +16,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Couldnt click Edit Profile button
 - Bug with routing when clicking on a project from a user profile and then trying to close it
-
 
 ## [2.2.1] - 2020-12-05
 ### Added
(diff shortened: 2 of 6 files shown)


src/graphql/queries/personQuery.js
```diff
@@ -33,6 +33,19 @@ export default
       neighborhood
       region
     }
+    events {
```
tibetsprague, line 36: once again im going to flag that i suspect loading all this at once is going to make loading user profiles too slow... im down to try it though

tibetsprague (commented):
same comment on backend, i think we should call these eventsAttending, since thats what it is. otherwise i might think its only events created by this user

src/graphql/queries/personQuery.js
lorenjohnson: I'm really not following your contention on this one. This particular query is only called once when a Member Profile is visited. The same data is needed for that page whether it's fetched via multiple graphql queries or one. This is not the query on `person` which is used on the member profile page (`community { members }` is used there without any of this additional data being queried).

tibetsprague: well can you explain to me why loading the Edit Profile page is now quite slow? I assume its because of the many queries happening on the back end that all have to complete before data is returned to the front-end. the way to make a web app feel fast is to load some data immediately and then load more using separate requests

tibetsprague: I mean performance a big rabbit hole to open up but its a real issue with Hylo right now that I promise is impacting adoption and happiness using it

tibetsprague: to be clear, I'm not asking for any change right now

lorenjohnson: Thanks for hashing this out a bit more @tibetsprague. In careful look at the Edit Profile page it appears that calling `fetchForCurrentUser` is querying much more than the data needing refresh for the edit profile (i.e. notifications, etc--as you pointed out). The edit page could be made available much more quickly in that case by querying just the fields on `Me` which are being edited instead of using the full `fetchForCurrentUser`, and also potentially taking-out the pending load for `savedSearches` across the whole UserSettings area and pushing it down into a loading indicator on the Edit Saved Searches tab.

* Copying this note onto a new ticket re. the loading time of Edit Profile.

wushenshen:
Linking to the discussion on the [node PR](https://github.com/Hylozoic/hylo-node/pull/566) - these events include both the ones the user created and the ones they are attending (when a user creates an event, they are added to the group that represents users attending)

lorenjohnson (approved):
👍

wushenshen:
@tibetsprague @lorenjohnson added the `eventsAttending` change to match the node PR

Issue #703, opened 2020-12-02
Clicking on a project on a profile while in a community should go to project in that community


tibetsprague:
Should always stay in the current context (public, all communities, network, community...) unless the project does not exist in the current context, then show it in the All Communities context

tibetsprague:
if it's easy lets also change how clicking on a post in the profile activity works to do the same thing. right now it just shows the post even if it does not exist in the current context

PR #705, opened 2020-12-03
Stay in current context when viewing a project from a profile

wushenshen:
Addresses #703 

This change makes it so the user stays in the current context, which I think is preferable for now. @lorenjohnson I think you were saying that the navigation was designed to stick with the current context, which is what I'm seeing here. 

The change to default to all communities instead of the current one if the project/post does not exist in the current context would involve [this navigation util function `baseUrl`](https://github.com/Hylozoic/hylo-evo/blob/dev/src/util/navigation.js#L60-L92).  It has to do with how the `communityUrl` is prioritized before the `defaultUrl`. `baseUrl` is used in a bunch of places though so it feels like a bigger conversation to have around what we want the UX around context to be.

### DIFF
--- a/src/routes/MemberProfile/MemberProfile.connector.js
+++ b/src/routes/MemberProfile/MemberProfile.connector.js
@@ -41,14 +41,12 @@ export function mapStateToProps (state, props) {
   }
 }
 
-export function mapDispatchToProps (dispatch, props) {
-  const { routeParams, querystringParams } = props
-
+export function mapDispatchToProps (dispatch) {
   return {
     fetchPerson: (id) => dispatch(fetchPerson(id)),
     blockUser: (id) => dispatch(blockUser(id)),
     push,
-    showDetails: (id) => dispatch(push(postUrl(id, routeParams, querystringParams)))
+    showDetails: (id, routeParams) => dispatch(push(postUrl(id, routeParams)))
   }
 }
 
--- a/src/routes/MemberProfile/MemberProfile.js
+++ b/src/routes/MemberProfile/MemberProfile.js
@@ -127,7 +127,7 @@ export default class MemberProfile extends React.Component {
           <div styleName='profile-subhead'>
             Projects
           </div>
-          {projects && projects.length > 0 && projects.map((p, index) => <Project key={index} memberCap={3} project={p} showDetails={showDetails} />)}
+          {projects && projects.length > 0 && projects.map((p, index) => <Project key={index} memberCap={3} project={p} routeParams={routeParams} showDetails={showDetails} />)}
         </div>
       </div>
       <div styleName='content'>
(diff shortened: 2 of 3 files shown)


tibetsprague:
yeah i agree I think this is good enough for now

src/routes/MemberProfile/MemberProfile.js
```diff
@@ -226,10 +226,11 @@ export function ActionDropdown ({ items }) {
     />
 }
 
-export function Project ({ memberCap, project, showDetails }) {
-  const { title, id, createdAt, creator, members } = project
+export function Project ({ memberCap, project, routeParams, showDetails }) {
+  const { title, id, communities, createdAt, creator, members } = project
+  const postTypeContext = communities.map(c => c.slug).includes(routeParams.slug) ? 'project' : undefined
```
tibetsprague, line 231: i dont really understand what this is doing. what happens if the project does not exist in the current communities?

tibetsprague: a comment would help

src/routes/MemberProfile/MemberProfile.connector.js
```diff
@@ -41,14 +41,12 @@ export function mapStateToProps (state, props) {
...
-export function mapDispatchToProps (dispatch, props) {
-  const { routeParams, querystringParams } = props
-
+export function mapDispatchToProps (dispatch) {
   return {
     fetchPerson: (id) => dispatch(fetchPerson(id)),
     blockUser: (id) => dispatch(blockUser(id)),
     push,
-    showDetails: (id) => dispatch(push(postUrl(id, routeParams, querystringParams)))
+    showDetails: (id, routeParams) => dispatch(push(postUrl(id, routeParams)))
```
tibetsprague, line 49: it feels weird to pass in the routeParams here, i dont quite get it...

lorenjohnson: The `util/navigation` helpers were designed to be coupled to the react router key schema in this way. We could de-couple them but in my experience we just end-up modeling or repeating the same context already represented appropriately by our routing. This may not be consistently applied throughout the navigation helpers, and if not, we should visit that at some point and do it consistently one way or another. You'll see in some connectors we pass `routeParams` and in others we break down routeParams into the various keys an then basically reconstruct the routeParams object when we generate a URL. There are important exceptions as in the cases where we don't want to reproduce the whole context on a link on purpose. In those cases we simply pass the keys we want to use to reproduce the link. Hope that makes sense. I obviously like the idea of passing around `routeParams` but I'm explaining this all more to make clear the direction and intent of the existing design more than to close any conversation about it :)

lorenjohnson (approved):
👍  LGTM! 

The UX of viewing a Member Profile and seeing the full feed of everything related to the user, without respecting the Community or Network context is how things are currently designed. The UX of switching context when clicking on a Project, or Post, that was not posted in the current Community or Network context would I think would actually be awkward. More awkward in my opinion than changing the top header context as you click on different items. So, all said, doing what is already done here does make the most sense to me, and I think the larger issue needs UI/UX consideration before we do anything differently.

tibetsprague:
ok i tried it, it works, lets do it, going to merge because im about to deploy the project stuff

PR #723, opened 2020-12-11
682 org affiliations

### LINKED ISSUE #682: Show organizational affiliations and community memberships on profile
New section on user profile that shows community memberships and custom organization affiliations. 

For the Hylo communities part, only show to the viewer the Hylo communities that they are also a member of, or that are public and have the setting to make visible their membership list.

https://www.figma.com/file/zAjzEqnfPjEU3zUUdBOrBI/Profile?node-id=0%3A1

Also, the Communities tab in the user settings now becomes Affiliations and has a new section for adding custom organizational affiliations

wushenshen:
Closes #682 

Related: https://github.com/Hylozoic/hylo-node/pull/574


tibetsprague:
Looks like some tests are failing

src/routes/MemberProfile/MemberProfile.js
```diff
@@ -131,6 +135,12 @@ export default class MemberProfile extends React.Component {
           </div>
           <SkillsToLearnSection personId={personId} editable={false} />
 
+          { memberships && memberships.length > 0 && <div styleName='profile-subhead'>Hylo Communities</div> }
+          { memberships && memberships.length > 0 && memberships.map((m, index) => <Membership key={m.id} index={index} membership={m} />) }
```
tibetsprague, line 139: i forget, does this do any checking on the back-end of whether the viewer is allowed to see this hylo community? im pretty sure we agreed that people can only see communities that the other person is a part of if it is public or if they too are a part of that community...

src/routes/UserSettings/CommunitySettingsTab/CommunitySettingsTab.js
```diff
@@ -1,51 +1,206 @@
...
-export function CommunityControl ({ membership, leaveCommunity }) {
-  const leave = () => {
-    if (window.confirm(`Are you sure you want to leave ${community.name}?`)) {
-      leaveCommunity(community.id)
-    }
+  deleteAffiliation = (affiliationId) => {
+    const { deleteAffiliation } = this.props
+    const { affiliations } = this.state
+
+    deleteAffiliation(affiliationId)
```
tibetsprague, line 76: i think this should still do a confirm dialog first

src/routes/UserSettings/CommunitySettingsTab/CommunitySettingsTab.js
```diff
@@ -1,51 +1,206 @@
...
+  }
+
+  toggleAddAffiliations = () => {
+    this.setState({ showAddAffiliations: !this.state.showAddAffiliations })
+  }
+}
+
+export function AddAffiliation ({ close, save }) {
+  const PREPOSITIONS = [
+    { id: 'of', label: 'of' },
```
tibetsprague, line 132: why do we need this if the id and label are always the same? imagining a future where they might not be? could just be an array for now

src/routes/UserSettings/CommunitySettingsTab/CommunitySettingsTab.js
```diff
@@ -1,51 +1,206 @@
...
+  ]
+
+  const [role, setRole] = useState('')
+  const [preposition, setPreposition] = useState(PREPOSITIONS[0].id)
+  const [orgName, setOrgName] = useState('')
+  const [url, setUrl] = useState('')
+
+  const canSave = role.length && orgName.length
+  
+  const URL_PROTOCOL = 'http://'
```
tibetsprague, line 144: should probably default to https

src/routes/UserSettings/UserSettings.connector.js
```diff
@@ -47,6 +48,7 @@ export const getMessageSettings = createSelector(
 
 export function mapStateToProps (state, props) {
   const currentUser = getMe(state, props)
+  const affiliations = currentUser.affiliations // TODO: How to make Affiliations a Redux Model?
```
tibetsprague, line 51: shall we pair on this?

tibetsprague: ok decided not to do this

src/routes/MemberProfile/MemberProfile.js
wushenshen: There's logic in the front-end to prevent the viewer from seeing the community if they are not part of it. Currently that is also filtering out public communities, I'm not sure why that is happening. Also in testing this I recreated the bug where a user's role shows up as "moderator" when the viewer is a moderator of that community.

wushenshen: Recap of our discussion in sprint planning - we agreed it works for the viewer to only see communities that they are also a member of (which is the functionality here). We should test whether the "moderator" bug appears in staging, as it could be a quirk of trying to log in as two different users locally.

tibetsprague (changes requested):
when testing i noticed a couple issues. you should be able to add an affiliation without a URL. and the error message when i tried to do so appeared way at the top of the page which i couldnt see, so errors should appear above or below the form itself. but actually we may not need any errors if you just make URL optional

wushenshen:
@tibetsprague Pushed a new commit to address your comments. It needs [this](https://github.com/Hylozoic/hylo-node/pull/581)

tibetsprague (approved):
nice. maybe take this moment to update broken snapshots from dev?

Issue #476, opened 2020-06-02
Hylo allows you to create empty comments


tibetsprague:
Just pressing enter on the comment box will do it, or space and enter. 

We should trim the text and only allow a comment to be posted if its not empty

PR #725, opened 2020-12-21
476 fix creating empty comments

veselinn:
Related to #476 

Adds a simple check if the content of the comment is empty. Displays an inline error if an empty comment is tried to be posted.
:warning: Please let me know if the error message is alright. Dunno how user friendly it is.

---

:rotating_light: This solution doesn't permit posting a comment with only an attachment. Attaching an attachment and pressing _enter_ triggers the empty comment error. This is because the attachment is at one level higher - in the [`CommentForm`](https://github.com/Hylozoic/hylo-evo/blob/15adf47506fa9604d5ef190678980aef2244f3d5/src/routes/PostDetail/Comments/CommentForm/CommentForm.js#L38-L53), while the fix is introduced in the `HyloEditor`.

**Q:** Why not introduce the fix in the `CommentsFrom#save`?
**A:** In the `save` we have the html content, an empty content is in the form of `<p></p>`. Hardcoding an empty check against `<p></p>` looks like a bad idea to me.(I have no good argument here, let me know if this approach should be pursued).

Let me know if this is a showstopper. Thanks for the opportunity!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+ - Doesn't allow posting of an empty comment.
 
 ## [2.2.3] - 2020-12-16
 ### Fixed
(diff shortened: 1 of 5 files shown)


tibetsprague (changes requested):
Thanks for this! I would actually say that we don't need an error message at all, since most people wouldn't expect to be able to post an empty comment. Also I think it is a dealbreaker to not be able to post a comment with just an attachement, that feels important to me

veselinn:
Happy new year! :sparkler: Thanks for the feedback!

 * Removed the error message.
 * Allows comments with no text but attachments.
 * Does not allow comment with no text and no attachments.
 
 Please, let me know if the solution is good enough. This was the best compromise I landed on.

tibetsprague (changes requested):
this looks good! only little change id ask for is to trim the text so to also reject comments that are just whitespace

veselinn:
Great call! Added a check for trimmed text.

tibetsprague (approved):

PR #732, opened 2020-12-30
511 community view switch

pospi:
Work in progress, untested in context switching from network <--> community.

### DIFF
--- a/src/util/navigation.js
+++ b/src/util/navigation.js
@@ -259,7 +259,23 @@ export function getViewContextInPath (pathname) {
   const match = matchPath(pathname, {
     path: '/(c|n)/:slug/:viewContext'
   })
-  return get('params.viewContext', match)
+  const match2 = matchPath(pathname, {
+    path: '/(all|public)/:viewContext'
+  })
+  const res = get('params.viewContext', match) || get('params.viewContext', match2)
+  switch (res) {
+    // route disjoint sub-paths back to parent section
+    case 'm':
+      return 'members'
+    // whitelist of possible view contexts; avoids :topicName being detected and 'topics' modal nav interfering
+    case 'map':
+    case 'project':
+    case 'event':
+    case 'members':
+      return res
+    default:
+      return undefined
+  }
 }
 
 export function gotoExternalUrl (url) {
--- a/src/util/navigation.js
+++ b/src/util/navigation.js
@@ -255,6 +255,13 @@ export function getNetworkSlugInPath (pathname) {
   return get('params.networkSlug', match)
 }
 
+export function getViewContextInPath (pathname) {
+  const match = matchPath(pathname, {
+    path: '/(c|n)/:slug/:viewContext'
+  })
+  return get('params.viewContext', match)
+}
+
 export function gotoExternalUrl (url) {
   return window.open(url, null, 'noopener,noreferrer')
 }
--- a/src/routes/AllTopics/AllTopics.js
+++ b/src/routes/AllTopics/AllTopics.js
@@ -99,7 +99,7 @@ export default class AllTopics extends Component {
     } = this.props
     const { totalTopicsCached } = this.state
 
-    return <FullPageModal fullWidth goToOnClose={`/${routeParams.context}/${routeParams.slug || ''}`}>
+    return <FullPageModal fullWidth>
       <div styleName='all-topics'>
         <div styleName='title'>{community ? community.name : network ? network.name : 'All'} Topics</div>
         <div styleName='subtitle'>{totalTopicsCached} Total Topics</div>
--- a/src/routes/AllTopics/AllTopics.js
+++ b/src/routes/AllTopics/AllTopics.js
@@ -99,7 +99,7 @@ export default class AllTopics extends Component {
     } = this.props
     const { totalTopicsCached } = this.state
 
-    return <FullPageModal fullWidth>
+    return <FullPageModal fullWidth goToOnClose={`/${routeParams.context}/${routeParams.slug || ''}`}>
       <div styleName='all-topics'>
         <div styleName='title'>{community ? community.name : network ? network.name : 'All'} Topics</div>
         <div styleName='subtitle'>{totalTopicsCached} Total Topics</div>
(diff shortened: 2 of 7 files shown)


lorenjohnson:
It seems like if we route matched the rendering of Drawer (and TopNav), from `route.match.params` those components would know then know what to render as well as have the needed bits to derive in-context URLs. Something like that can back us out of these repeated "top of the tree" (PrimaryLayout) path-to-regexp helpers...  Wholeheartedly agree with you that we should basically be able to generate a URL from a predictable and react-router defined set of parameters (i.e. context, slug, id, entity, action, entityId, viewContext, etc). I don't think we're actually far off from that, but you'll see a lot of legacy code here before there were URL helpers or even this version of react-router being used. Mostly the actual state is reflected in the URL throughout the app, with the big exception of maps and a couple smaller areas.

Anyway--I haven't QA'd it, but the code looks just fine.

It looks like a CI/testing run didn't get kicked-off on the PR. Have you been able to get the tests running? Are all the tests passing?

pospi:
Haven't run locally yet, but looks like they are passing now. I should probably add one around the new URL helper at least, I can see some of the others have tests.

pospi:
Added some tests for the URL helper and whitelisted what is considered a "view context" to avoid the nav stack issues with topic pages. Ready to review & merge!

src/util/navigation.js
```diff
@@ -255,6 +255,29 @@ export function getNetworkSlugInPath (pathname) {
   return get('params.networkSlug', match)
 }
 
+export function getViewContextInPath (pathname) {
```
tibetsprague, line 258: i think this is ok for now, but it feels like a better solution might be to add the :view route parameter to all the routes in PrimaryLayout (we already do this for all the map routes), then we dont need to keep this helper function updated as we change routes

tibetsprague (approved):
i think this is a good enough solution for now, though i left a comment about a potentially better way. also one thing we didnt discuss is what should happen when clicking on a community somewhere in the app that is not the navigation drawer, like in a post where it lists all the communities that the post was shared to. i think probably in that case we do actually want to switch to the home page, but we should probably discuss that

src/util/navigation.js
lorenjohnson: Was saying the same above, though more verbosely :) I don't want to hold things up, but I do think it'd be best--and for all the reasons you're citing Pospi--to make this adjustment now. 

Also agree that the other inline links back to community should be considered. I just did the diligence on mobile for this and it'd be nice if a reliable things happened across both apps. We could create a new ticket for that if it feels like this one is scope creeping.

pospi:
I think I get you... so in `primaryLayout` all the routes that we want matched as a view would be annotated with that parameter, meaning it could be pulled out of the redux state instead of the URL?

It does seem like it would require some fairly scary reauthoring of that file, because all the `POST_TYPE_CONTEXT_MATCH` stuff would have to be realigned onto `:viewContext`, and there are extensive places where that is used. Totally agree we don't want to do that right now.

Before closing this- should I log a new issue to track the final polished solution? Or are such changes likely to be superceded by upgrading React Router at some point anyway?

lorenjohnson:
> I think I get you... so in `primaryLayout` all the routes that we want matched as a view would be annotated with that parameter, meaning it could be pulled out of the redux state instead of the URL?
> 
> It does seem like it would require some fairly scary reauthoring of that file, because all the `POST_TYPE_CONTEXT_MATCH` stuff would have to be realigned onto `:viewContext`, and there are extensive places where that is used. Totally agree we don't want to do that right now.
> 
> Before closing this- should I log a new issue to track the final polished solution? Or are such changes likely to be superceded by upgrading React Router at some point anyway?

The solution I was imagining would use the URL helpers more than changing a lot about routes. If in all the places where we want to respect the groupContext (community or network) context we send the routeParams to a helper we should be able to generate a link to the right context. 

I am not certain this would require changing any routes. As a couple examples: 

* `<Drawer styleName={cx('drawer', { hidden: !isDrawerOpen })} {...{ community, network, viewContext }} />` once route matched instead of directly rendered would receive `routeParams` which currently would include all the `context`, `communitySlug` (simply called `slug` for now) and `networkSlug`. The `communityUrl` helper would then take `routeParams` as its parameter (instead of only `slug` currently) and use that to generate the correct in context link for the links in `CommunityRow`

This is somewhat of an inversion of your solution, but I do think it's something we should do now as, given where we're headed in the next 2 months, getting this solution "right" (within reasonable limits) seems of greater importance than getting the functionality there today. 

Hope that makes sense and that I'm not missing something big. I apologize for not thinking this through more in my first set of notes.

lorenjohnson:
> > I think I get you... so in `primaryLayout` all the routes that we want matched as a view would be annotated with that parameter, meaning it could be pulled out of the redux state instead of the URL?
> > It does seem like it would require some fairly scary reauthoring of that file, because all the `POST_TYPE_CONTEXT_MATCH` stuff would have to be realigned onto `:viewContext`, and there are extensive places where that is used. Totally agree we don't want to do that right now.
> > Before closing this- should I log a new issue to track the final polished solution? Or are such changes likely to be superceded by upgrading React Router at some point anyway?
> 
> The solution I was imagining would use the URL helpers more than changing a lot about routes. If in all the places where we want to respect context (currently community or network) we send the `routeParams` to the URL helper we should be able to generate links in the right context.
> 
> I am not certain this would require changing any routes. For example:
> 
> * `<Drawer styleName={cx('drawer', { hidden: !isDrawerOpen })} {...{ community, network, viewContext }} />` once route matched instead of directly rendered would receive `route.match.params` which currently would include all the `context`, `communitySlug` (simply called `slug` for now) and `networkSlug`. The `communityUrl` helper would then take `routeParams` as its parameter (instead of only `slug` currently) and use that to generate the correct in context link for the links in `CommunityRow`
> 
> This is somewhat of an inversion of your solution, but I do think it's something we should do now. Given where we're headed in the next 2 months getting this solution "right"--within reasonable limits--is of greater importance than getting the functionality there today.
> 
> Hope that makes sense and that I'm not missing something big. I apologize for not thinking this through more in my first set of notes.

Here is a rough and dirty of the sort of the direction. This works except I didn't mend the public|all context to work with this (it could be normalized), AND as you'll see I've essentially copy pasted the current routing map to get all the right match params. That route map (`routesWithDrawer`) could be shortened to just the routes that keep context, and whatever we do with it I think this same map could be used for both `Drawer` and `Navigation` (where we're currently using a `routesWithNavigation`).

If we decide to not go this way now we can link back to this branch and to this PR for notes. I do however think it's best to get going in this direction now though, what we're about to do with more abstract group contexts is going to only intensify our needs for good context handling:

https://github.com/Hylozoic/hylo-evo/tree/511-keep-context-prototype

pospi:
Canceling this branch- new version reduces complexity of the routing system.

PR #739, opened 2021-01-11
Update node to v12 and redux orm to the latest

tibetsprague:


lorenjohnson (approved):
👍  LGTM. I have minimally QA'd this, but I am comfortable with what I'm seeing here likely not blowing anything up. I would like to upgrade redux (and related dependencies) shortly too.

PR #742, opened 2021-01-11
Holonic architecture

tibetsprague:


lorenjohnson (approved):
I really want to say something interesting here, but basically.... LGTM 👍  I will say: I really like that when you did this you went all the way with the renaming and didn't leave any "community" or "network" naming behind. I mean, of course, but it could have been tempting to leave behind some vestiges which could have left things muddy for later devs...

tibetsprague:
whoop thanks! now want to look at the backend one ;)

PR #752, opened 2021-01-19
Upgrade to node v15, etc

lorenjohnson:
Upgrades the following dependencies to the latest versions including necessary adjustments:

* Redux 3.x < 4.x
* React Router 4.x > 5.x
* Jest v23 > v26 (including enzyme, etc)
* Node v12 >  v15
* heroku-buildpack-nodejs to latest, reapplying our server side rendering patch (see https://github.com/Hylozoic/heroku-buildpack-nodejs/tree/run-webpack-build-v182). This was required to support later versions of node. 

Adds:

* React Testing Library (see https://reactjs.org/docs/testing.html#tools and https://testing-library.com/docs/react-testing-library/intro). This is not currently utilized in our tests, but is available for use on future tests.


config/webpack.config.prod.js
```diff
@@ -2,7 +2,7 @@ const path = require('path')
 const webpack = require('webpack')
 const HtmlWebpackPlugin = require('html-webpack-plugin')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
-const ManifestPlugin = require('webpack-manifest-plugin')
+const { WebpackManifestPlugin } = require('webpack-manifest-plugin');
```
tibetsprague, line 5: remove semicolon

src/routes/Messages/MessageSection/MessageSection.test.js
```diff
@@ -55,105 +55,29 @@ const messages = [
 let wrapper, instance, socket
 
 beforeEach(() => {
-  socket = { on: jest.fn() }
-})
-
-it('renders as expected', () => {
```
tibetsprague, line 61: should we file a ticket to redo these tests?

tibetsprague (approved):
do you think we should try to roll this out now with the bug fixes that are lined up?

lorenjohnson:
> do you think we should try to roll this out now with the bug fixes that are lined up?

#772 is based off oof this branch, so yeah👍🏼

src/routes/Messages/MessageSection/MessageSection.test.js
lorenjohnson: Meh. Honestly, I don't know that I'd bother in this case.

Issue #747, opened 2021-01-13
Clicking on a topic in All Topics view doesn't go anywhere


lorenjohnson:
This was a bug introduced in https://github.com/Hylozoic/hylo-evo/pull/738 such that the `.../topics` :view param is retained when navigating to a topic. Should be a quick fix. 

A nice to have here would be to just move from `.../:topicName` to `.../:view(topics)/:topicName` making adjustments accordingly (don't forget notifications and notification emails, etc. Possibly also a legacy redirect).

Issue #744, opened 2021-01-13
Trying to go to a non existent topic in a community should go to a blank feed


tibetsprague:
Same view as going to a topic with no posts

Issue #706, opened 2020-12-03
Edit profile page takes too long to load - 15 seconds + for Tibet


lorenjohnson:

lorenjohnson:
Thanks for hashing this out a bit more @tibetsprague. In careful look at the Edit Profile page it appears that calling fetchForCurrentUser is querying much more than the data needing refresh for the edit profile (i.e. notifications, etc--as you pointed out). The edit page could be made available much more quickly in that case by re-querying just the fields on Me which are being edited instead of using the full fetchForCurrentUser, and also potentially taking-out the pending load for savedSearches across the whole UserSettings area and pushing it down into a loading indicator on the Edit Saved Searches tab.

Copying this note onto a new ticket re. the loading time of Edit Profile.

PR #772, opened 2021-01-27
Edit User Profile loading, All Topics > Topic nav, Topic feed for topic not found 

lorenjohnson:
Fixes for #706, #747, and #744


src/routes/UserSettings/UserSettings.js
```diff
@@ -12,20 +11,7 @@ import FullPageModal from 'routes/FullPageModal'
...
-
 export default class UserSettings extends Component {
-  static propTypes = {
-    currentUser: object,
-    onClose: func
-  }
-
-  componentDidMount () {
-    const { currentUser } = this.props
-    this.props.fetchForCurrentUser()
```
tibetsprague, line 25: we dont need this even if we just go straight to the user settings URL?

tibetsprague (approved):
code looks good, havent tested yet, one question

src/routes/UserSettings/UserSettings.js
lorenjohnson: Correct, and tested. It is reliably handled by the auth logic on the way in. Ah, actually, I forgot. It likely can/should be removed here in CommunitySettings as well https://github.com/Hylozoic/hylo-evo/blob/50a634204b8e3b1e4ca9713306edfa7c63e4f385/src/routes/UserSettings/CommunitySettingsTab/CommunitySettingsTab.js#L38. I'll do that shortly and include it in this branch.

tibetsprague: i guess the only reason to have these is if there is some change that happens that has not propagated to the front-end yet for some reasons so then looking at these tabs is showing out of date information. im not sure if that is a really valid concern. But for example if someone has two tabs open and makes a change in one and then goes to the settings page in the other.

tibetsprague: thoughts? even if we did keep them they should not block loading of the UI

lorenjohnson: That is why they would have existed yes, and I've finger tested the cases I could come-up with which could potentially generate that condition and didn't ever find a problem. Regardless, I think its' fair to say that this SHOULDN'T ever happen if we have faith in our Redux ORM use and if we employ it consistent throughout the app. If it does happen it'd be better to catch it as a bug in this area than in more subtle places, and again I don't think there is are issues in this case. 

Finally, to load  this here "just in case" while not block editing has the result of updating  the component while it's being edited overwriting any not yet saved edits. We could of course build around that in some way but that would add mess for what I think in this case is no benefit.

lorenjohnson: Also keep in mind that the things are edited on the Profile screen aren't available to change anywhere within the app outside of this screen. This is true with affiliations as well, and for memberships I think it's reliable as it's something we more fundamentally can't afford to have out of sync elsewhere in the app... Make sense?

lorenjohnson: To be fair there is at least one edge case: if you have multiple Hylo windows open and/or mobile and you're making edits in this are on both at the same time without reloading. I think that is a tolerable vulnerability with low consequences.

tibetsprague: yeah I hear all that. im good with this

PR #775, opened 2021-01-27
736 threaded comments

pospi:
Central changes to handle comment threading on the client implemented and polished.

Depends on https://github.com/Hylozoic/hylo-node/pull/595

Still a little left to do to paginate backwards through long sub-comment threads, but a UX review now would be great!


src/components/HyloEditor/HyloEditor.js
```diff
@@ -81,9 +84,19 @@ export default class HyloEditor extends Component {
...
       this.setState(this.defaultState(this.props))
     }
+
+    // handle initial focus behaviour
+    if (this.props.focusOnRender !== prevProps.focusOnRender) {
+      this.setState({ didInitialFocus: false })
+    }
+    if (this.props.focusOnRender && !this.state.didInitialFocus) {
+      this.focus()
+      this.setState({ didInitialFocus: true })
```
tibetsprague, line 98: this doesnt quite seem right to me, you could do two re-renders here with two setState calls, seems like you should check to only do one between both these conditions?

src/css/global/fonts.scss
```diff
@@ -15,11 +15,11 @@
 
 @font-face {
   font-family: 'hylo-evo-icons';
-  src:  url('/assets/fonts/icons/hylo-evo-icons.eot');
-  src:  url('/assets/fonts/icons/hylo-evo-icons.eot#iefix') format('embedded-opentype'),
-    url('/assets/fonts/icons/hylo-evo-icons.ttf') format('truetype'),
-    url('/assets/fonts/icons/hylo-evo-icons.woff') format('woff'),
-    url('/assets/fonts/icons/hylo-evo-icons.svg#hylo-evo-icons') format('svg');
+  src:  url('/assets/fonts/icons/hylo-evo-icons.eot?ne5jg1');
```
tibetsprague, line 18: ooh, good call, we should always do this to cache bust

src/graphql/fragments/postFieldsFragment.js
```diff
@@ -1,3 +1,24 @@
+// :TODO: clean this up and use proper query fragments?
```
tibetsprague, line 1: this seems fine to me, not quite sure the advantages of real fragments

tibetsprague (commented):
looks good, will do some more testing

src/css/global/fonts.scss
lorenjohnson: Hmm... This is handled automatically by the CSS/PostCSS loader through the WebPack build and if for some reason that is not working should be remedied.

src/graphql/fragments/postFieldsFragment.js
lorenjohnson: Yeah, it's a discussion but I know of no material advantages today in our stack for using real fragments vs JS wrapped ones in this way other than explicit graphql syntax validation. However, there is as of the last time I checked one big disadvantage to using native graphql (fragments or othewise) in our setup which is that `.graphql` files don't recompile within the dev server so whenever adding or updating them you need to restart the dev server (`yarn start`) to have the changes. That gotcha can ruin an afternoon if you're not aware of it.

tibetsprague (changes requested):
it all looks good, except i am seeing a bug where i click to load 36 previous comments and then it shows 16 previous comments, then it click it again and it says click to show -4 previous comments, and then i click it again and it says click to show -24 previous comments, and then i click it and it is showing all comments

tibetsprague (changes requested):
another bug im seeing is i can only ever see the last two sub comments under a comment. when i add a third comment the first one disappears. even after page reload

src/components/HyloEditor/HyloEditor.js
pospi: Nice catch! Let me know if da5c6b97 doesn't resolve that for you.

src/css/global/fonts.scss
pospi: Ok took it out in 46e50383

src/graphql/fragments/postFieldsFragment.js
pospi: Yeah, I found that one already. All good!

pospi:
> it all looks good, except i am seeing a bug where i click to load 36 previous comments and then it shows 16 previous comments, then it click it again and it says click to show -4 previous comments, and then i click it again and it says click to show -24 previous comments, and then i click it and it is showing all comments

Oh, well that's kinda weird but I sortof put all bets off on the pagination logic until the issues with the backend returning the wrong totals are resolved. I made myself a note to revisit and retest once the issue has been fixed in `holonic-architecture`.

pospi:
> another bug im seeing is i can only ever see the last two sub comments under a comment. when i add a third comment the first one disappears. even after page reload

This should be fixed now, it was to do with the logic for only showing a subset from the server. I am now adjusting totals for the newly injected comments which I think is probably the simplest/best way of handling this in this particular UI. But if the totals cached in Redux need to also be updated, there's more work there.

Probably worth testing what happens when somebody else adds a comment while you're looking at a thread, because that might throw totals off and require deeper integration than what I have added in 1203743d.

tibetsprague (changes requested):
another issue: normally comments you create have 3 dots next to them that allow you to edit or delete the comment. not seeing those for sub-comments you create

tibetsprague:
also a reminder to change the base of these PRs to the holonic-architecture branches

Issue #754, opened 2021-01-20
Navigating to most Member Profiles crashes with ReduxORM Membership query error (on dev)


lorenjohnson:
@tibetsprague I found after merging the ReduxORM branch to dev. Navigating to `/c/hylo/m/982` (or I think maybe most any member profile) results in `Error: Expected to find a single row in `Membership.get`. Found 2.. 

I looked into it a bit and didn't come-up with a ready solution. I may by the time I finish this update but I wanted to alert you to it in case you have a ready answer. This relates to `Model.extension.js#safeGet` which we previously wrapped in a `try / catch` and might have been throwing away the error, though it does look like a new version ReduxORM issue.

tibetsprague:
Seems like a good thing i removed that try catch as it was swallowing real errors. I think this error was why the red dot counts are not being reset currently on prod...?

Issue #735, opened 2021-01-03
Viewing a member profile from all communities view is sometimes broken


tibetsprague:
https://www.hylo.com/m/19644 gives error: TypeError: Cannot read property 'eventsAttending' of null

PR #779, opened 2021-02-01
Fix crashing when viewing a member profile

tibetsprague:
Also, this might fix resetting unread post counts?
And fixes when trying to view profile of member you aren't allowed to see

Fixes https://github.com/Hylozoic/hylo-evo/issues/754
Fixes https://github.com/Hylozoic/hylo-evo/issues/735


lorenjohnson (changes requested):
Again-- ahhh yes, this is going to set a couple things straight!  It seems that `src/graphql/fragments/meQueryFragment.js` is not used anywhere, doesn't hold anything that is not represented in `MeCompleteFragment` and should be removed. Thoughts?

tibetsprague:
@lorenjohnson huh thats true, ill remove it!

Issue #757, opened 2021-01-20
Compact stream view


tibetsprague:

pospi:
Very close now with a little cleanup afterwards that @brodeur may do as part of this task or in another-

- [ ] Finesse the border underneath the view controls (it doubles up on the card view presently)
- [ ] Implement the left sidebar shadow on all views
- [ ] Decide on appropriate visual separator between the list view and member sidebar on wider screens

Still for me to do before submitting an initial PR-

- [ ] Add icons
- [ ] Add tooltips to view mode buttons
- [ ] Figure out if we should render links in body copy in the preview text. Note this may be hard to wire up all the needed events.

PR #799, opened 2021-02-16
757 compact stream view

pospi:
This is a work-in-progress; adds the bulk of the code for the new compact stream view.

Assigning to @brodeur as there are a few stylistic concerns to tie off- see https://github.com/Hylozoic/hylo-evo/issues/757#issuecomment-777912339.

Please also note the refactoring I've done to support better consistency in tooltip styles- comment tooltips are the first to be swapped out for the standard appearance in 78eb25ef. If you'd like to do an audit of these and perhaps create some style variants of the new `Tooltip` component, please do- you can start by searching all occurrences of `ReactTooltip`.


src/components/PostListRow/PostListRow.js
```diff
@@ -0,0 +1,74 @@
...
+    details,
+    creator,
+    createdAt,
+    commentersTotal,
+    votesTotal,
+    myVote,
+    topics
+  } = post
+
+  if (!creator) { // PostCard guards against this, so it must be important? ;P
```
tibetsprague, line 30: hah, no idea

src/components/PostListRow/PostListRow.js
```diff
@@ -0,0 +1,74 @@
...
+
+  const creatorUrl = personUrl(creator.id, routeParams.slug)
+  const numOtherCommentors = commentersTotal - 1
+  const unread = false
+
+  return (
+    <div styleName={cx('post-row', { unread, expanded })} onClick={showDetails}>
+      <div styleName='votes'>
+        <a onClick={voteOnPost} styleName={cx('vote-button', { voted: myVote })}
+          data-tip-disable={myVote} data-tip='Upvote this post so more people see it.' data-for='postfooter-tt'>
```
tibetsprague, line 42: not always seeing this tooltip for some reason

src/components/PostListRow/PostListRow.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const creatorUrl = personUrl(creator.id, routeParams.slug)
+  const numOtherCommentors = commentersTotal - 1
+  const unread = false
+
+  return (
+    <div styleName={cx('post-row', { unread, expanded })} onClick={showDetails}>
+      <div styleName='votes'>
+        <a onClick={voteOnPost} styleName={cx('vote-button', { voted: myVote })}
```
tibetsprague, line 41: it does feel weird to have the post open when you upvote it to me

src/components/Tooltip/Tooltip.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import ReactTooltip from 'react-tooltip'
+import './Tooltip.scss'
+
+const Tooltip = (props) => {
```
tibetsprague, line 5: nice!

tibetsprague (commented):
this is looking sexy! thanks

src/components/PostListRow/PostListRow.js
pospi: Fixed, I was trying to reuse an existing ID but it's not always rendered!

src/components/PostListRow/PostListRow.js
pospi: That is a weird event bubbling bug I am not sure about :/

pospi: Fixed in ebf67e60

pospi:
Fixed up remaining rough edges per @tibetsprague's review, integrated with new settings fields for automatic persistence (see https://github.com/Hylozoic/hylo-node/pull/623).

tibetsprague (approved):
looks good! seems like we have to wait until the landing page is integrated before adding in the stream?

tibetsprague:
@pospi this is looking awesome! I am seeing some issues with the remembering functionality is that supposed to be fully working? storing the last settings in the database and using them on load of the stream?

pospi:
They *should* be- are you running the backend branch as well?

May differ slightly from the spec, reading back through. I have added the saving logic into the settings switching UI- so, if you follow a link to the stream with different settings to your saved view will be overridden- but only until you navigate away from & return to the stream, after which you should see based on your preferences again. If you change the view settings via the linked view then they just get saved as normal and loaded as the new defaults.

tibetsprague:
im not seeing the save working. can we look at together?

PR #802, opened 2021-02-17
768 groups joining groups

tibetsprague:


src/store/actions/groupActions.js
```diff
@@ -0,0 +1,207 @@
+import { get } from 'lodash/fp'
```
lorenjohnson, line 1: I think it's helpful to sequester all of these into one file, but I would call it `groupRelationshipActions.js` vs `groupActions.js` as there are other group related actions with their own files.

src/store/reducers/ormReducer/index.js
```diff
@@ -274,6 +296,18 @@ export default function ormReducer (state = {}, action) {
       comment.delete()
       break
 
+    case DELETE_GROUP_RELATIONSHIP: {
+      if (payload.data.deleteGroupRelationship.success) {
+        const gr = GroupRelationship.safeGet({ parentGroup: meta.parentId, childGroup: meta.childId })
+        if (gr) {
+          gr.delete()
+          clearCacheFor(Group, meta.parentId)
```
lorenjohnson, line 304: Did these `clearCacheFor`s still prove to be necessary after you solved the ordering issue otherwise? I expect so, but just checking. I was really hoping that the recent ReduxORM update would have made this sorta hack unnecessary.

src/routes/Groups/Groups.store.js
```diff
@@ -1,60 +1,19 @@
...
+import getMyModeratedGroups from 'store/selectors/getMyModeratedGroups'
+import getRouteParam from 'store/selectors/getRouteParam'
+
+export const getPossibleRelatedGroups = ormCreateSelector(
+  orm,
+  getGroupForCurrentRoute,
+  (session, props) => getCurrentlyRelatedGroupIds(session, { groupSlug: getRouteParam('groupSlug', session, props) }),
+  getMyModeratedGroups,
+  (session, group, currentRelationships, moderatedGroups) => {
+    // TODO: check for cycles, cant add a grandparent as a child
```
lorenjohnson, line 14: Assuming that this case isn't covered otherwise, what happens if someone does that?

src/routes/Groups/Groups.js
```diff
@@ -1,125 +1,110 @@
...
-
-const { array } = PropTypes
-
-const sortOptions = [
-  { id: 'name', label: 'Alphabetical' },
-  { id: 'num_members', label: 'Popular' },
-  { id: 'created_at', label: 'Newest' }
-]
-
 export default class Groups extends Component {
```
lorenjohnson, line 12: This would make a tidy functional component with a `useState()` for the invite picker... In lock-step with React moves I try and make new components that can be (which is most) functional components using hooks. It hurts a little at first, but starts to feel good soon after. Put here for discussion, not thinking you'd change this component here now.

lorenjohnson (approved):
👍  All looks good with one minor quibble about naming of `groupActions` which seems worth maybe amending, and a couple of notes for discussion but which shouldn't be blocking to merging. The semantics of groups joining groups gets a little strange, doesn't it? Either way good work!

lorenjohnson:
> 👍 All looks good with one minor quibble about naming of `groupActions` which seems worth maybe amending, and a couple of notes for discussion but which shouldn't be blocking to merging. The semantics of groups joining groups gets a little strange, doesn't it? Either way good work!

Actually, forgot to mention that the tests/snapshots seem to be out of date. Also, you might want to merge in that small update from `holonic-architecture` which fixes the analytics event name.

src/routes/Groups/Groups.js
tibetsprague: i know, i too want to move more towards functional components, there's just some momentum i have towards what i know a bit better

src/routes/Groups/Groups.store.js
tibetsprague: well, the main issue that might happen which i havent yet looked at is that when looking at a parent group you see all the posts of children groups you are also a member of so we would need to check for the cycle there. maybe its not actually a big deal if there are cycles... just strange and maybe indicating something weird going on

src/store/reducers/ormReducer/index.js
tibetsprague: ill double check

PR #813, opened 2021-02-24
791 user join group questions

tibetsprague:
- settings for join questions on the admin group settings page
- seeing and answers questions when going to join a group
- seeing answers in the join requests admin page


lorenjohnson (approved):
👍 Yep, that all looks just fine.

PR #864, opened 2021-04-06
863 new posts appear in stream immediately

tibetsprague:
Plus other little fixes when creating posts
- When in public context default to is public being true and stay in public stream after creating post
- When opening and closing the create/edit post modal keep the stream query params

Fixes: https://github.com/Hylozoic/hylo-evo/issues/863
Fixes: https://github.com/Hylozoic/hylo-evo/issues/862

### DIFF
--- a/src/components/CreateModal/CreateModal.connector.js
+++ b/src/components/CreateModal/CreateModal.connector.js
@@ -1,18 +1,20 @@
 import { connect } from 'react-redux'
 import { push } from 'connected-react-router'
 import { get, omit } from 'lodash/fp'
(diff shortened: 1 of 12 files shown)


lorenjohnson (approved):
LGTM 👍  So, if I've understood this correctly in the case of #863 then it was not getting the result to the cached query results after create because the query string params were not carried over?

tibetsprague:
No that issue was handled entirely by this commit https://github.com/Hylozoic/hylo-evo/pull/864/commits/0c28a96fea6736da84f930be73f01339df07b8da

The query Param stuff was fixing other bugs 

Sent by m

PR #873, opened 2021-04-14
groups network map visualization

cpolitano:


src/components/GroupNetworkMap/GroupNetworkMap.js
```diff
@@ -0,0 +1,27 @@
...
+  const nodesData = networkData.nodes
+
+  if (linksData.length && nodesData.length) {
+    const containerRef = React.useRef(null)
+
+    React.useEffect(() => {
+      let destroyFn
+
+      if (containerRef.current) {
+        const { destroy } = runForceGraph(containerRef.current, linksData, nodesData)
```
tibetsprague, line 16: seems like this entire function could just be 
`if (containerRef.current) {
return runForceGraph(containerRef.current, linksData, nodesData)
}
return null
`

tibetsprague (approved):
wow that does look complicated! i do wonder about using a library but the nice thing about this is we are in total control of everything. looks good to me

Issue #827, opened 2021-03-16
Autocomplete topics when starting with a #


tibetsprague:
When typing in new topics for a post you should be able to start by typing a # and still have it autocomplete existing topics. The # should be ignored so that #food should be the same as the food topic

PR #887, opened 2021-04-29
Made it so you can type in topics with hashtags. Small bug(?) introdu…

brodeur:
…ced... when entering a hashtag, if the only thing entered is the octothorpe, the status message still says "Start by adding a topic" but it is now a selectable item in the list. Pressing return cancels the input.

What's happening is that AsyncCreatableSelect thinks there's an input (which there is, even though we take out the octothorpe) and so it's calling formatOptionLabel on line 141.

The behavior should, at some point, shift to the default behavior of noOptionsMessage on line 138. To do this, I think we have to capture the input before it goes to AsyncCreatableSelect, not quite sure how to do this.

As it stands, this behavior is much better than current behavior and the "bug" is not all that bad.

### DIFF
--- a/src/components/TopicSelector/TopicSelector.js
+++ b/src/components/TopicSelector/TopicSelector.js
@@ -78,6 +78,9 @@ export default class TopicSelector extends Component {
   }
 
   handleInputChange = async (input) => {
+    if (input.charAt(0) === '#') {
+      input = input.slice(1)
+    }
     this.setState({ input })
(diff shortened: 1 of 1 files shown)


brodeur:
The bug is hardly noticeable as a bug

tibetsprague:
Fixes https://github.com/Hylozoic/hylo-evo/issues/827

tibetsprague (approved):

PR #902, opened 2021-06-14
Show the timezone for events' start and end times

tibetsprague:
Using moment-timezone for now, would like to switch away from moment eventually to Luxon or date-fns
I am also removing unused moment locales in this commit too


tibetsprague:
Fixes https://github.com/Hylozoic/hylo-evo/issues/904

PR #931, opened 2021-08-13
Husky integration: Added lint check on pre-commit hook and test check on pre-push hook

thomasgwatson:
Yo,

While waiting on context for other tickets, I jumped into this. Thought it would be useful.

PR includes:
- linting updates to put whole project into compliance with linter
- addition of pre-commit lint check (sub second check if no errors)
- addition of test check on push to remote (this takes a bit longer. Do we want to give it a go and then remove if its being too much of a pain?)

Note: this should act independent of our IDEs and can also be skipped by adding `--no-verify`

Open questions:
- Are there any times CircleCI or something else automated makes a commit or push? If we want to avoid those triggering the hooks, we just need to set `HUSKY=0` in their envs


tibetsprague (approved):
Lets give this a shot!

PR #934, opened 2021-08-23
Hylo.com links target _self

### LINKED ISSUE #922: Links to posts in other posts should open in same tab, not a new one

thomasgwatson:
Closes #922 

@lorenjohnson , turns out there was already a `clickCatcher` munging links on the components anyway...

I put the hylo.com regex in the navigation folder. It should capture any hylo with http or https, but not email addresses or `something.hylo.com`. Included a link to the regex101 for testing purposes.

### DIFF
--- a/src/components/ClickCatcher/ClickCatcher.js
+++ b/src/components/ClickCatcher/ClickCatcher.js
@@ -1,6 +1,6 @@
 import PropTypes from 'prop-types'
 import React from 'react'
-import { hyloUrlRegex } from 'util/navigation'
+import { HYLO_URL_REGEX } from 'util/navigation'
 
 export default function ClickCatcher ({ tag, handleMouseOver, navigate, ...props }) {
   if (!['div', 'span', 'p'].includes(tag)) {
@@ -21,7 +21,7 @@ export default function ClickCatcher ({ tag, handleMouseOver, navigate, ...props
       node.setAttribute('target', '_blank')
     }
 
-    const matches = [...node.getAttribute('href').matchAll(hyloUrlRegex)]
+    const matches = [...node.getAttribute('href').matchAll(HYLO_URL_REGEX)]
     if (matches[0] && matches[0].length === 2) {
       node.setAttribute('target', '_self')
       const urlPath = matches[0][1] === '' ? '/' : matches[0][1]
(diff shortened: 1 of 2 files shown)


src/util/navigation.js
```diff
@@ -13,6 +13,7 @@ export const REQUIRED_EDIT_POST_MATCH = `:detail(post)/:postId(${POST_ID_MATCH})
 
 export const GROUP_DETAIL_MATCH = `:detail(group)/:detailGroupSlug`
 export const OPTIONAL_GROUP_MATCH = `:detail(group)?/(:detailGroupSlug)?`
+export const hyloUrlRegex = /http[s]?:\/\/hylo\.com(.*)/g // https://regex101.com/r/yQxli1/1/
```
tibetsprague, line 16: i think we want to use the getHost() function here so we can use/test locally and on staging. will that work with subdomains like staging.hylo.com? lets figure that out

src/components/ClickCatcher/ClickCatcher.js
```diff
@@ -19,6 +20,13 @@ export default function ClickCatcher ({ tag, handleMouseOver, navigate, ...props
     if (node.getAttribute('target') !== '_blank') {
       node.setAttribute('target', '_blank')
     }
+
+    const matches = [...node.getAttribute('href').matchAll(hyloUrlRegex)]
+    if (matches[0] && matches[0].length === 2) {
+      node.setAttribute('target', '_self')
+      const urlPath = matches[0][1] === '' ? '/' : matches[0][1]
+      node.setAttribute('href', urlPath)
```
thomasgwatson, line 28: for any hylo.com links, this link just cuts off the host altogether. So the `href` is just the path and will just hit that path on whatever host (local, staging or prod) the application is running on.

Anticipating that this will work on local, staging and prod

src/util/navigation.js
thomasgwatson: I'm anticipating this will work on each env, without the use of `getHost()`.

Added a note here:
https://github.com/Hylozoic/hylo-evo/pull/934/files#r695211734

lorenjohnson: It's a quibble but wondering what not `HYLO_URL_REGEX` to match how we otherwise name constants?

lorenjohnson (approved):
The one quibble, otherwise LGTM👍🏼 Glad this worked-out to be so rather tidy and straightforward!

src/util/navigation.js
thomasgwatson: Ah yes! It only got moved there at the last moment and I missed that detail

Issue #912, opened 2021-07-30
Hide top navigation bar in iframes


tibetsprague:
For now when embedding Hylo in an iFrame we hide the top navigation so people cant navigate to a different context. This is for Coffee Shop but may make sense for everyone

tibetsprague:
this is not needed right now because we can do layoutFlags=mobileSettings and it hides the nav!

PR #955, opened 2021-10-14
Hide top navigation bar when in an iFrame

tibetsprague:
TODO: decide whether we also want to hide the search, chat, notifications and profile links

Fixes https://github.com/Hylozoic/hylo-evo/issues/912


tibetsprague:
no longer needed because layoutFlags=mobileSettings now does exactly this from the URL!

PR #979, opened 2021-11-01
Evo 679 coordinates as location inputs

thomasgwatson:
OKAY....

This is basically a draft because the test issues are driving me batty. Pushing it up so we can trouble-shoot collectively (I have `.skip' called the tests that were giving me grief.)

I have tried to mock an `async` call directly and then done the dependency injection (passing it in via props), mocked it and in both instances it breaks the tests. Could there be something about `async` keyword that the babel compiler for the test suite doesn't like? Am I messing it up in some fashion?

### DIFF
--- a/src/components/PostCard/PostTitle/PostTitle.js
+++ b/src/components/PostCard/PostTitle/PostTitle.js
@@ -17,8 +17,10 @@ export default function PostTitle ({
   if (locationObject) {
     if (locationObject.addressNumber !== null && locationObject.addressNumber !== '') {
       generalLocation = `${locationObject.addressNumber} ${locationObject.addressStreet}, ${locationObject.city}, ${locationObject.region}`
-    } else {
+    } else if (locationObject.city !== null && locationObject.city !== '') {
       generalLocation = `${locationObject.city}, ${locationObject.region}`
+    } else {
+      generalLocation = locationObject.fullText
     }
   }
 
--- a/src/components/PostEditor/PostEditor.js
+++ b/src/components/PostEditor/PostEditor.js
@@ -25,6 +25,7 @@ import SendAnnouncementModal from 'components/SendAnnouncementModal'
 import PublicToggle from 'components/PublicToggle'
 import styles from './PostEditor.scss'
 import { PROJECT_CONTRIBUTIONS } from 'config/featureFlags'
+import { ensureLocationIdIfCoordinate } from 'components/LocationInput/LocationInput.store'
 
 export const MAX_TITLE_LENGTH = 50
 
@@ -48,7 +49,8 @@ export default class PostEditor extends React.Component {
     post: PropTypes.shape(POST_PROP_TYPES),
     removeLinkPreview: PropTypes.func,
     titlePlaceholderForPostType: PropTypes.object,
-    updatePost: PropTypes.func
+    updatePost: PropTypes.func,
+    fetchLocation: PropTypes.func
   }
 
   static defaultProps = {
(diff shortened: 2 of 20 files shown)


thomasgwatson:
#679 notionally closes this ticket

src/components/PostCard/PostTitle/PostTitle.js
```diff
@@ -17,8 +17,10 @@ export default function PostTitle ({
   if (locationObject) {
     if (locationObject.addressNumber !== null && locationObject.addressNumber !== '') {
       generalLocation = `${locationObject.addressNumber} ${locationObject.addressStreet}, ${locationObject.city}, ${locationObject.region}`
-    } else {
+    } else if (locationObject.city !== null && locationObject.city !== '') {
```
tibetsprague, line 20: city both times, should be region

src/components/PostEditor/PostEditor.js
```diff
@@ -327,26 +328,28 @@ export default class PostEditor extends React.Component {
...
       endTime, location, locationId, isPublic
     } = this.state.post
+    let coordLocationId = null
     const details = this.editor.current.getContentHTML()
     const topicNames = this.topicSelector.current.getSelected().map(t => t.name)
     const memberIds = members && members.map(m => m.id)
     const eventInviteeIds = eventInvitations && eventInvitations.map(m => m.id)
     const imageUrls = imageAttachments && imageAttachments.map(attachment => attachment.url)
     const fileUrls = fileAttachments && fileAttachments.map(attachment => attachment.url)
+    coordLocationId = await ensureLocationIdIfCoordinate({ fetchLocation, location, locationId })
```
tibetsprague, line 349: this feels smelly, shouldnt all the location stuff happen in the polling so that by the time you save its all figured out?

src/routes/GroupSettings/GroupSettingsTab/GroupSettingsTab.js
```diff
@@ -92,9 +94,15 @@ export default class GroupSettingsTab extends Component {
...
     this.updateSetting(key, changed)({ target: { value } })
 
-  save = () => {
+  save = async () => {
     this.setState({ changed: false })
-    this.props.updateGroupSettings(this.state.edits)
+    const { group, fetchLocation } = this.props
+    let coordinateLocationId
+    if (group && this.state.edits.location !== group.location) {
+      coordinateLocationId = await ensureLocationIdIfCoordinate({ fetchLocation, location: this.state.edits.location, locationId: this.state.edits.locationId })
```
tibetsprague, line 102: again why not do this all in the polling? otherwise we are blocking save with another server call

src/util/geo.js
```diff
@@ -9,3 +10,40 @@ export function locationObjectToViewport (priorViewport, locationObject) {
...
+  }
+  return { coordinate: { lng: parsedCoordinate.getLongitude(), lat: parsedCoordinate.getLatitude(), string: coordinate }, error }
+}
+
+export function convertCoordinateToLocation (coordinate) {
+  let city = ''
+
+  let addressNumber = ''
+  let addressStreet = ''
+  console.log(coordinate, 'cooridnate inside')
```
tibetsprague, line 31: remove console

src/util/geo.js
```diff
@@ -9,3 +10,40 @@ export function locationObjectToViewport (priorViewport, locationObject) {
...
+  try {
+    parsedCoordinate = new CoordinateParser(coordinate)
+  } catch (err) {
+    error = err
+    return { coordinate: null, error }
+  }
+  return { coordinate: { lng: parsedCoordinate.getLongitude(), lat: parsedCoordinate.getLatitude(), string: coordinate }, error }
+}
+
+export function convertCoordinateToLocation (coordinate) {
```
tibetsprague, line 26: add tests for these functions

tibetsprague (commented):
seems like a good one to pair on, i think theres a better way

src/components/PostCard/PostTitle/PostTitle.js
thomasgwatson: You have to check both for a empty string and null. Elsewhere the check is done for both city and region, functionally one just needs to check one of them to get the desired result

src/components/PostEditor/PostEditor.js
thomasgwatson: `pollingFetchLocation` is only triggered if you select an option from the autocomplete drop-down. Coordinates pasted or typed in will not provide a result to select.  So coordinates need to be handled in another way.

If we offer the fullText as a selectable option in the autocomplete list, in the coordinate case, the user might not select the coordinate fullText anyway. And in the non-coordinate case, we might give the user the impression that an address is correct (and shows up in the geocoder) when it doesn't actually exist.

src/components/PostCard/PostTitle/PostTitle.js
tibetsprague: ah ok, we use isEmpty from lodash in many places

thomasgwatson:
Closed because I need to push recent changes on a different branch

Issue #981, opened 2021-11-02
Only 2 child comments showing at a time


tibetsprague:
No way to load additional comments

PR #982, opened 2021-11-02
Fix bug that only shows last 2 child comments on a comment thread

tibetsprague:
Fixes https://github.com/Hylozoic/hylo-evo/issues/981


thomasgwatson (approved):
lgtm

PR #993, opened 2021-11-08
EVO-857 - Display timezone for datetimes in events, requests, offers and resources

Sofacosta:


src/components/PostCard/PostHeader/PostHeader.js
```diff
@@ -9,24 +9,25 @@ import Highlight from 'components/Highlight'
...
 import { filter, isFunction, isEmpty } from 'lodash'
 import cx from 'classnames'
 
-const formatStartDate = (startTime) => {
+const formatStartDate = (startTime, from) => {
   const current = moment()
   let start = ''
   if (moment(startTime).isAfter(current)) {
-    start = moment(startTime).format('MMM D YYYY')
+    start = from
```
Sofacosta, line 22: I use passed-in formatted start/end dates instead of format them here. formatDatePair was used in events so I imported it so it can be used in requests, offers and resources.

src/components/PostCard/PostHeader/PostHeader.js
```diff
@@ -91,9 +92,12 @@ export default class PostHeader extends PureComponent {
 
     const typesWithTimes = ['offer', 'request', 'resource', 'project']
     const canHaveTimes = typesWithTimes.includes(type)
+
+    const { from, to } = formatDatePair(startTime, endTime, true)
+    const startDate = (startTime && from) && formatStartDate(startTime, from)
+    const endDate = (endTime && to) && formatEndDate(endTime, to)
```
Sofacosta, line 98: Here we check the values and then pass in formatted start/end dates with timezone

src/util/index.js
```diff
@@ -19,7 +19,7 @@ export function isPromise (value) {
 
 export const inflectedTotal = (word, count) => `${count.toLocaleString()} ${inflection.inflect(word, count)}`
 
-export const formatDatePair = (startTime, endTime) => {
+export const formatDatePair = (startTime, endTime, returnAsObj) => {
```
Sofacosta, line 22: I am modifying `formatDatePair` to accept an optional parameter, `returnAsObj`.

src/util/index.js
```diff
@@ -37,17 +37,18 @@ export const formatDatePair = (startTime, endTime) => {
 
   if (endTime) {
     if (end.year() !== start.year()) {
-      to = end.format(' - ddd, MMM D, YYY [at] h:mmA z')
+      to = end.format('ddd, MMM D, YYYY [at] h:mmA z')
```
Sofacosta, line 40: fixes a bug that caused year to render as a 6-digit number

src/util/index.js
```diff
@@ -37,17 +37,18 @@ export const formatDatePair = (startTime, endTime) => {
...
-      to = end.format(' - ddd, MMM D [at] h:mmA z')
+      to = end.format('ddd, MMM D [at] h:mmA z')
     } else if (end.date() !== start.date()) {
-      to = end.format(' - ddd, MMM D [at] h:mmA z')
+      to = end.format('ddd, MMM D [at] h:mmA z')
     } else {
-      to = end.format(' - h:mmA z')
+      to = end.format('h:mmA z')
     }
+    to = returnAsObj ? to : ' - ' + to
```
Sofacosta, line 48: make sure to keep the correct formatting for `Events` which is also using this function

src/util/index.js
```diff
@@ -37,17 +37,18 @@ export const formatDatePair = (startTime, endTime) => {
...
   }
 
-  return from + to
+  return returnAsObj ? { from, to } : from + to
```
Sofacosta, line 51: when `returnAsObj` is true, return `from` and `to` in an object instead of a concatenated string

src/components/PostCard/PostHeader/PostHeader.js
```diff
@@ -9,24 +9,25 @@ import Highlight from 'components/Highlight'
...
 import Icon from 'components/Icon'
 import { personUrl, topicUrl } from 'util/navigation'
+import { formatDatePair } from 'util/index'
 import { humanDate } from 'hylo-utils/text'
 import './PostHeader.scss'
 import { filter, isFunction, isEmpty } from 'lodash'
 import cx from 'classnames'
 
-const formatStartDate = (startTime) => {
+const formatStartDate = (startTime, from) => {
```
tibetsprague, line 18: so this function is not really doing anything anymore, except checking whether the start time is in the future. you should either remove the function, or change it to be a generic one that checks if a date is in the future

src/components/PostCard/PostHeader/PostHeader.js
tibetsprague: again i think you can remove the formatStartDate and formatEndDate functions as they are no longer really doing anything

tibetsprague (changes requested):
Thanks Sofia! A few comments but its close!

src/components/PostCard/PostHeader/__snapshots__/PostHeader.test.js.snap
```diff
@@ -163,7 +163,9 @@ exports[`matches announcement snapshot 2`] = `
   </div>
   <div
     data-stylename="timeWindow"
-  />
+  >
+    request ended 
+  </div>
```
Sofacosta, line 168: These show up because I removed the possibility that end date is an empty string.

src/util/__snapshots__/index.test.js.snap
```diff
@@ -2,10 +2,10 @@
 
 exports[`formatDatePair displays differences of dates 1`] = `"Mon, Feb 4, 2019 at 6:41PM GMT"`;
 
-exports[`formatDatePair displays differences of dates 2`] = `"Mon, Feb 4, 2019 at 6:41PM GMT - 9:41PM GMT"`;
+exports[`formatDatePair displays differences of dates 2`] = `"Mon, Feb 4, 2019 at 6:41PM GMT - Mon, Feb 4 at 9:41PM GMT"`;
```
Sofacosta, line 5: This is the result of simplifying the end date logic in formatDatePair. It will always render a complete date.

tibetsprague: not totally convinced this is better, but im open to convincing :) to me if the date stays the same then you just want to know the time, not display the same date twice

Sofacosta: My intent here was to provide clarity as the formatting didn't make it obvious to me that the event was starting and ending on the same day. However, if the preference is to revert it to the current formatting I can push a fix to this PR. No problem :)

tibetsprague: Really what it should be if the event starts and ends on the same is like this: Mon, Feb 4, 2019 at 6:41PM - 9:41PM GMT. So back to the way it was but only show the time zone once. This makes it more obvious that its a time range on the same day. Let's do that. Thanks!

Sofacosta: I made the changes requested, let me know what you think :)

tibetsprague (approved):
beautiful, thanks!

PR #999, opened 2021-11-24
New registration flow with email verification

tibetsprague:
Enter email first, get sent email with verification code.
Enter code or click on link in email to verify
Finish registration by entering name and password
Refactor TextInput component to handle internal labels
Going to /signup or /login when logged in will redirect to the app instead of letting you try to sign up again
Validate email when logging in
Add password confirmation to signup flow


tibetsprague:
fixes https://github.com/Hylozoic/hylo-evo/issues/997

lorenjohnson:
Testing this now and a question: 

I'm not yet clear on the necessity of the cookie storage of the verification information. 

Since we do store both the email and verification code in the `user_verification_codes` table it seems like I should be able to still verify my email address even if I no longer have the cookie (referencing the `created_at` on that table to determine expiry). 

Currently when I test in a private window without the cookie I get an error in dev, I'm thinking that there should at least be an error page there describing that the verification code is invalid/expire and that the verification step needs to repeated once again?

Thoughts?

lorenjohnson:
> 

This is also of concern to me if someone initiates the email verification step on a computer and then receives and clicks on the link in the email from their phone...

tibetsprague:
@lorenjohnson hmm yeah this is something i struggled with. I was trying to figure out how to pass the verified email along to the final step of registration in evo. I wanted to track which stage the person was at and add another layer of making sure they are who they are, but you're right that they might want to open the link in another browser or device. Unlikely, but possible. I could just pass the email through the URL or the redux state but then they would have to start over on a refresh, which is i guess ok. Hmm, lets talk it through

tibetsprague:
also can you describe how exactly you get the error? it should definitely say link expired or something

lorenjohnson:
Sending to screen casts to demonstrate the error (in dev mode) when entering an invalid code and clicking "verify" despite the error message presented. In the private window demo note that the user will get a "this code is invalid" error even though it's not, in the case of just entering the code wrong, ignoring the error message, and clicking "Verify Email" anyway we should probably at minimum deactivate the button until the code is validated on the front end. 

I don't feel it is that unlikely that someone would start the process in one browser and validate in another for a few reasons:

1) Their may be browser someone is using to sign-up vs the the default browser one may have set in their email client
2) The new email notification may first appear on someone's phone and they may click on it their first, because humans are chaotic :)
3) The user may get past entering the email address then come back later (on a different device) when they get around to clicking on the link in the notification email.

In any case I'm not yet getting the dilemma you're having, but I may just not be getting it. As it is it seems to me that an XHR call to validate the email address and confirm it's validated addresses the above concerns and generally is a more standard/durable and (debatably) secure solution. 

https://www.icloud.com/iclouddrive/0LYMNxZ1aHmv0mBdxyLEoJC3g#Enter_code_in_new_browser_without_cookie

https://www.icloud.com/iclouddrive/08c-ZmfAQ757PiAiWIR4VBwjQ#Enter_wrong_code_and_submit

lorenjohnson:
https://user-images.githubusercontent.com/6294/144116623-a3e72c5e-fa9a-411f-86b0-10b43d974689.mov


https://user-images.githubusercontent.com/6294/144116765-2bc86d5c-d4b8-4522-8c97-42bcb4a8a82c.mov

src/util/index.js
```diff
@@ -62,3 +62,9 @@ export function hexToRgb (hex) {
 export function inIframe () {
   return window.location !== window.parent.location
 }
+
+/* eslint-disable */
+export const validateEmail = email => {
```
lorenjohnson, line 67: Good on you for tidying that up 👍🏼

lorenjohnson (commented):
That all looks good; consider this comment : https://github.com/Hylozoic/hylo-node/pull/707#discussion_r767329536

lorenjohnson (approved):
👍🏼

PR #1004, opened 2021-12-03
956 map icons on safari

tibetsprague:
Fix map icons not appearing on safari

Update all our map-gl and deck-gl code to the latest

Fixes https://github.com/Hylozoic/hylo-evo/issues/956


thomasgwatson (approved):
lgtm

Issue #978, opened 2021-10-27
When more than 6 attachments on a post you cant see them all


tibetsprague:
Because they are horizontal and go off the screen. We could list them vertically, or in a grid.

PR #1016, opened 2021-12-15
Display attachments in a column instead of a row so you can see more

tibetsprague:
Fixes https://github.com/Hylozoic/hylo-evo/issues/978


thomasgwatson (approved):
lgtm

Issue #921, opened 2021-08-06
Issues with welcome message on explore page


tibetsprague:
- [ ] Editing the welcome message does not update it in real time
- [ ] Hyperlinks and newlines don't work in the message, make sure links can be clicked on and newlines actually create newlines in the message

This code is at the LandingPage.js route component, and the WelcomeWidget.js

PR #1017, opened 2021-12-15
Fix issue where editing welcome widget message wouldn't update immedi…

tibetsprague:
…ately

Also make it so links work in the welcome message, and you can use markdown to style the text!

Fixes https://github.com/Hylozoic/hylo-evo/issues/921

### DIFF
--- a/src/components/Widget/WelcomeWidget/WelcomeWidget.js
+++ b/src/components/Widget/WelcomeWidget/WelcomeWidget.js
@@ -1,3 +1,4 @@
+import { markdown } from 'hylo-utils/text'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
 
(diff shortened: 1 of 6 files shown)


src/components/Widget/WelcomeWidget/WelcomeWidget.js
```diff
@@ -12,10 +13,11 @@ export default class WelcomeWidget extends Component {
 
   render () {
     const { group, settings = {} } = this.props
+    const message = settings.text || `We're happy you're here! Please take a moment to explore this page to see what's alive in our group. Introduce yourself by clicking Create to post a Discussion sharing who you are and what brings you to our group. And don't forget to fill out your profile - so likeminded new friends can connect with you!`
     return (
       <div styleName='welcome-widget'>
         <h2>{settings.title || `Welcome to ${group.name}!`}</h2>
-        <p>{settings.text || `We're happy you're here! Please take a moment to explore this page to see what's alive in our group. Introduce yourself by clicking Create to post a Discussion sharing who you are and what brings you to our group. And don't forget to fill out your profile - so likeminded new friends can connect with you!`}</p>
+        <p dangerouslySetInnerHTML={{ __html: markdown(message) }} />
```
thomasgwatson, line 20: Are there alternatives to dangerouslySet, I thought they were pushing peeps away from using that?

tibetsprague:
Hmm, probably. I just copied this from what we do in the group sidebar about text right now.

thomasgwatson:
Ahh ok... well, something for the backlog :)

thomasgwatson (approved):

tibetsprague:
@thomasgwatson actually i think this is something we should deal with. we are calling the markdown function from hylo-utils/text and it doesnt do any sanitizing! so i a going to add that there: https://github.com/Hylozoic/hylo-utils/blob/master/src/text.js

PR #1021, opened 2021-12-16
Make delete and deactivate modal responsive

### LINKED ISSUE #1019: Make delete/deactivate modal responsive for mobile
Modal is currently flowing pass the screen width and some of the copy is obscured

thomasgwatson:
closes #1019

### DIFF
--- a/src/components/ModalDialog/ModalDialog.js
+++ b/src/components/ModalDialog/ModalDialog.js
@@ -51,7 +51,7 @@ export default class ModalDialog extends Component {
 
   static defaultProps = {
     closeOnSubmit: true,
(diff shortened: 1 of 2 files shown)


tibetsprague (approved):
looks fine but make sure you test with other ModalDialogs that this doesnt mess up their styling, especially since you increased the max-width

thomasgwatson:
> looks fine but make sure you test with other ModalDialogs that this doesnt mess up their styling, especially since you increased the max-width

Yep, great, I checked each of the use-cases. Most passed in their own max-width and the one that doesn't is fine. Merging

PR #1027, opened 2021-12-27
1025 incorrectly showing posts as ended

### LINKED ISSUE #1024: When clicking the three vertical dots on a person's profile on chrome using my computer, the box to block the user is not visible.
bug

Please see image. The box is hidden and needs to be moved further left

### LINKED ISSUE #1025: Excluding dates on resource/project/offer/event request shows the post as "ended"
These posts show up in the stream as 'ended'. Some posts don't have an expiry date because they are relevant to the group for the long-term. 

If user doesn't chose dates, it shows as 'resource' ended

tibetsprague:
Fix issue where posts without a start/end time would always show as ENDED
Fixes #1025
Also cleanup display of post start and end times to be easier to read - don't uppercase text, better syntax.
Also if a post was "completed" before it ended show the end date as the date it was completed
Also tweak styling of post headers to look better - better positioning for spacer before announcement icon, and less whitespace below post header, calendar date doesnt overlap post header on mobile

Improve copy for comment box placeholder 
"Add a comment..." instead of reusing same copy from post box which is What's on your mind?
Geoff asked for this

Make sure you can see the dropdown to block a member 
Fixes #1024

### DIFF
--- a/src/components/CommentCard/CommentCard.connector.js
+++ b/src/components/CommentCard/CommentCard.connector.js
@@ -3,9 +3,7 @@ import { push } from 'connected-react-router'
 import { postUrl } from 'util/navigation'
 
 export function mapStateToProps (state, props) {
-  return {
-    routeParams: props.routeParams || {}
-  }
+  return { }
 }
 
 export function mapDispatchToProps (dispatch, props) {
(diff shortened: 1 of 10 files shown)


thomasgwatson (approved):
lgtm

PR #1032, opened 2022-01-07
Changes related to Map WebView integration on Mobile

lorenjohnson:
This branch will be deployed to staging for staging-based Mobile releases for testing.


src/routes/MapExplorer/MapExplorer.js
```diff
@@ -64,7 +67,11 @@ export default class MapExplorer extends React.Component {
   }
 
   componentDidMount () {
+    const mobileSettingsLayout = this.context
+    this.setState({ mobileSettingsLayout })
```
tibetsprague, line 71: why do you need this in the state? are you going to change it?

lorenjohnson: Good point. I'd mis-read something that recommended this way of accessing the context value, but on review it's unnecessary. Will just use a `const mobileSettingsLayout = this.context` directly in render

lorenjohnson: Uhm actually this is a bug anyway, it should have been `const { mobileSettingsLayout } = this.context`--ode to code reviews. Thanks for doing it!

tibetsprague (approved):

Issue #1018, opened 2021-12-15
Port widget data selection and fetching to hooks


thomasgwatson:
Widgets largely rely on parent components to pass down their state. This increases the complexity of using them across different parts of the codebase.

This ticket is aiming to clean up how widget data is fetched and selected out of the redux store

PR #1072, opened 2022-01-20
Evo 1018 port widget data selection

thomasgwatson:
So I want to propose a pattern for us to follow.

As I mention in the readme update: 
_A new pattern being introduced for the application is the addition of `useEnsureX`, which is a custom hook that handles both selecting `X` from the store and, if it isn't present, also handles the fetching of that data._

Take a look at the below initial example `useEnsureCurrentGroup` and tell me what you think about the idea of moving to this idea. In my eyes, the biggest benefit is that when you drop one of these hooks into a component, you know that component won't have an implicit dependency on something else to make the request for the data it needs.


src/hooks/useEnsureCurrentGroup/index.js
```diff
@@ -0,0 +1,29 @@
...
+import fetchGroup from 'store/actions/fetchGroupDetails'
+import { useSelector, useDispatch } from 'react-redux'
+import presentGroup from 'store/presenters/presentGroup'
+import getGroupForCurrentRoute from 'store/selectors/getGroupForCurrentRoute'
+import { useRouter } from 'hooks/useRouter'
+import { createSelector } from 'reselect'
+import isPendingFor from 'store/selectors/isPendingFor'
+import { FETCH_GROUP_DETAILS } from 'store/constants'
+
+const selectAndPresentGroup = createSelector(
```
thomasgwatson, line 11: Added an extra reselect here to ensure `presentGroup` wasn't being re-run on every pass through of a component using this hook, and thus possibly forcing unnecessary re-rendering if the result of this hook was passed down to a child component

thomasgwatson:
I had greater ambitions to replace `getWidgetItems` and push those data selections/transformations into their respective widget components but a bunch of the logic around visibility is handled in the `Widget` component. So I don't currently have a good idea about how to handle that...

thomasgwatson:
Closes #1018

src/components/Widget/Widget.js
```diff
@@ -20,6 +20,11 @@ import PrivacyWidget from './PrivacyWidget'
 import RichTextWidget from './RichTextWidget'
 import JoinWidget from './JoinWidget'
 import TopicsWidget from './TopicsWidget'
+import { useSelector, useDispatch } from 'react-redux'
+import { useEnsureCurrentGroup } from 'hooks/useEnsureCurrentGroup'
```
lorenjohnson, line 24: Why not default export?

src/hooks/useEnsureCurrentGroup/index.js
```diff
@@ -0,0 +1,29 @@
...
+  (state, router) => getGroupForCurrentRoute(state, router),
+  (group) => presentGroup(group)
+)
+
+export function useEnsureCurrentGroup () {
```
lorenjohnson, line 16: Why not the default export?

lorenjohnson (approved):
I don't have the context of having looked at the `farm` branch yet so can't comment on a lot of this. I think the `useEnsureCurrentGroup` hook is great and likely something we'll want to use (sic) more generally once this is merged. As for generalizing on the `useEnsureX` pattern for other areas I'm hesitant because of the things I've named which I've also wadded into just now. Great work though, keep going! And I still think we should take an hour or two and jam/mindmeld/spitball some of this together soon.

tibetsprague (approved):
cool!

Issue #909, opened 2021-07-14
Improve message UI/UX


tibetsprague:
- [x] make message box more obvious in chat interface - hard to see at the bottom with very little contrast
- [x] Make it clearer when/how to choose a person and then start the conversation - e.g. once you select the people hit go (press enter?) and then it clears the list and opens up message view? Right now pressing tab is the way to do it (or clicking on the message box) and maybe this is good enough
- [x] After you send the first message to the chat the focus should stay in the chat box (autoFocus)
- [x] Arrow key doesn't work when new person box is blank
- [ ] Scroll down as you move down list with arrow keys (i think not for now, created some weirdness, FB doesnt do it)
- [x] Make it more obvious when focus is in message box (mostly done, work with aaron)
- [x] Messages start at the bottom 
- [x] Changing view between new messages and message list - change transitions to work correctly with React router
- [x] Style issues with bottom menu
- [x] When tabbing out of choose person area the dropdown of people should disappear
- [x] Are we now calling it Messages or Conversations?

PR #1073, opened 2022-01-21
Chat improvements

tibetsprague:
So many improvements!
Fixes https://github.com/Hylozoic/hylo-evo/issues/909

- make message box more obvious in chat interface - hard to see at the bottom with very little contrast
- Make it clearer when/how to choose a person and then start the conversation 
- After you send the first message to the chat the focus should stay in the chat box (autoFocus)
- Arrow key works when new person box is blank
- Make it more obvious when focus is in message box
- Messages start at the bottom
- Changing view between new messages and message list works correctly on mobile with nice CSS transitions: Added "inbox" route at /messages that is used on mobile to show list of all messages.
- While loading messages view show Messages header immediately instead of weird All Groups context header
- Add hover CSS in thread list
- Make it more obvious which current thread is being viewed
- Clean up code and switch to function components
- Fix style issues with bottom menu
- When tabbing out of choose person area the dropdown of people disappears
- Update react-transition-group to the latest


src/components/Pillbox/__snapshots__/Pillbox.test.js.snap
```diff
@@ -7,32 +7,57 @@ exports[`Pillbox renders 1`] = `
...
+    <TransitionGroup
+      childFactory={[Function]}
+      component="div"
     >
-      <Pill
-        id={1}
+      <CSSTransition
+        classNames={
+          Object {
+            "enter": undefined,
```
thomasgwatson, line 17: Is this how you disable some of the functionality of `CSSTransition`?

thomasgwatson (approved):
lgtm

src/components/Pillbox/__snapshots__/Pillbox.test.js.snap
tibetsprague: im not sure why the snapshot shows undefined actually. something to do with how css modules work in snapshots?

PR #1078, opened 2022-01-22
SASS and Node dependency upgrades and some attention to WebPack bundle

lorenjohnson:
This branch does 5 things. 2 bigger things to note, and 3 smaller things. Mostly dependency related stuff related to the load time performance dive (i.e. the removal of `faker` from our bundle). Generally this all stuff that may reduce a keep us out of some trouble going forward:

* **Replaces `node-sass` with the current drop-in replacement `sass` (Dart-SASS)**
`node-sass` is deprecated and a well known trouble maker, not just for us. Dart-SASS is also a whole lot faster and the actively maintained. The `sass` library is somewhat more strict in it's enforcement of the SASS/SCSS syntax standard, but fortunately it only required changing one overly-ambitious CSS selector to get us current. * On boot of the app you'll see 4-5 deprecation warnings arising from out-of-date `bootstrap` library code we're still depending on, but it does and doesn't break anything.

* **Upgrades to Node 16**
Doing this here as well to consolidate testing efforts. `heroku-18` supports node 16 and it seems on my M1 that despite `nvm` appearing to do an arm64 build from source for node 15 it ends-up installing the x86 version in the end. Node 16 for building our app using arm64 speeds-up my dev server build times by 2x and may have some effect some build/deploy time improvement on Heroku. The local build time speed increase really helps when testing server front-end build, bundle, and server related changes. Also, for whatever reason in recent work I've encountered 2 dependencies which I couldn't bring to current version due to being compatible with either Node 12-14 or Node 16, but not node 15. By all accounts it seems just better and doesn't appear to change or break anything.

* **Removes dependency on `samplePost` (deprecated) from `PostCard` which depended on `faker`**
We're still using `samplePost` (renamed to `fakePost`) in a test but removing the dependency from `PostCard` means that WebPack is no longer including `faker` in a our production bundle effectively reducing the initial unpacked bundle size by a bit over 1 MB! Total size of unpacked bundles goes from ~7.5M to ~6MB. *Also updated the `faker` dependency to `@faker-js/faker` which seems to be the now official repo after the original authors recent stunt.*

* **Adds `BundleAnalyzerPlugin` to WebPack and related `yarn analyze` command**
This WebPack plugin calculates and stores in depth bundle analysis after each production build. The data is accessible locally by running `yarn build` and then `yarn analyze`. Super helpful to have this on-hand for future load time considerations as we continue adding features and growing the codebase. 

* **WebPack config change**
Added a line to our `SplitChunks` config to accommodate potential Code Splitting in the near future. This change doesn't have any impact on the current bundles. 

* **Moves recent `devDependencies` additions back into the `dependencies` in `package.json`**
WebPack handles the bundling and pays no regard to this delineation. There is nothing else in our toolchain which looks to or relies upon this separation as far as I've known, and I've looked for it. In earlier days we leaned into this fact and encouraged awareness of it by not making delineation. It's fairly common practice and I think still a good idea as it's one less thing to think about.

Been testing this both locally as well as on staging (it should be current there, it's auto-deploying this branch atm) and have not found any issues.


thomasgwatson (approved):
lgtm

tibetsprague (approved):
amazing!

PR #1095, opened 2022-02-01
Evo-1075 move focus to message input when pressing enter 

Sofacosta:


tibetsprague (commented):
actually before i approve can you test one thing on this branch, after removing the autoFocus, if you click on an existing message thread does it automatically put focus on the message box? lets make sure that still works, and if it doesn't you may need to do something new to ensure it still works

Sofacosta:
> actually before i approve can you test one thing on this branch, after removing the autoFocus, if you click on an existing message thread does it automatically put focus on the message box? lets make sure that still works, and if it doesn't you may need to do something new to ensure it still works

I tested it on my browser and it works. Should I write a test for this behavior?

tibetsprague:
That would be a hard test to write, I think we’re good for now!

tibetsprague (approved):

Issue #1093, opened 2022-02-01
When searching for locations, locations do not always appear; error in console


brodeur:
When searching for locations on the map, the locations populate in the dropdown correctly but I am not always able to zoom to the location. See attached screenshot, which implies lat/lng is not being passed to map correctly

brodeur:
<img width="1792" alt="Screen Shot 2022-02-01 at 9 55 47 AM" src="https://user-images.githubusercontent.com/83412/152024110-c60494af-cb22-4699-a9a5-548a9f99c98a.png">

Issue #606, opened 2020-08-16
Have map URL and parameters always stay in sync


tibetsprague:
Map URL should include the current location (lat, lng?), and all the filter params so it can be copied and pasted. Changing the URL should also update the map.

Issue #616, opened 2020-08-25
Load posts for map much more quickly


tibetsprague:
Initial load of posts doesn't need all post data just type and name and location. 

Then have separate query to load posts for the drawer and load as you scroll

This means we will also need to use server side filtering for the drawer posts

Question: how many posts should we load on the map at first? eventually we want to load all, but that might require some server side caching or tile creation or something... maybe we can try loading all of them if we are just loading the very basic post info for the map

- [ ] Load more than first 100 posts, how many?

wushenshen:
Now that the map drawer is expanded by default more than just the type, name, and location are needed. If the map drawer stays out as the viewport changes, same situation. Everything in the PostCard is needed for the posts in the viewport - so looks like the attachments, communities, isPublic, creator, timestamps, etc.

tibetsprague:
not for the map, only for the drawer. that is the point of this ticket, only load the extra stuff in a separate query in the drawer, and only for the first 10 posts or so so it will be fast. for the map we want to load all the posts so we need minimal data to be faster

PR #1096, opened 2022-02-04
616 map load performance and much more

tibetsprague:
Faster load time for the map data, load less data for every post and group on the map so we can load more of them from the server more quickly. Then load the extra data needed for posts in in the drawer separately.
This means drawer posts are now paginated and reload from the server anytime a parameter or filter changes on the front-end
https://github.com/Hylozoic/hylo-evo/issues/616

Show members and groups in the drawer in separate tab from posts, alphabetically ordered for now
Fixes https://github.com/Hylozoic/hylo-evo/issues/669

Don't show fulfilled posts on the map
Add Groups to the feature toggles so you can can turn them on or off
You can now sort posts by post date alongside latest activity and popular
Add ability to replace query results cache with new data instead of appending to it 
Add longer delay/debounce to loading map content when changing the viewport
Wait 800ms instead of 300ms so it doesn't constantly try to load new content as you scroll around or zoom in and out. This creates a smoother experience and a lot less load on the servers.
Stop passing lastPostId to saved searches since back-end figures this out now
Export GroupCard from GroupWidget to use in map drawer
Reorder post types to be discussion, request, offer, resource, project, event (members and groups after)
In create modal and map features dialog

Sync map state to URL so you can copy, paste and bookmark
Keep current map view and filters if you browse to other view in the same group and then come back.
Fixes https://github.com/Hylozoic/hylo-evo/issues/606

Don't show toggle for Members on the public context

Fix bug that prevented searching for and jumping to a point location (like an address) (fixes: https://github.com/Hylozoic/hylo-evo/issues/1093)


src/components/Widget/GroupsWidget/GroupsWidget.js
```diff
@@ -72,3 +54,26 @@ export default class GroupsWidget extends Component {
     )
   }
 }
+
+export function GroupCard ({ group, routeParams, className }) {
```
thomasgwatson, line 58: Ehh, Aaron said this version of the groupcard will probably be overwritten once we settle on the new version from the Group Explorer :(

thomasgwatson: Does this version now show up in the context of the map?

tibetsprague:
Yeah thats what I decided to use in the map drawer for now (at Aaron’s suggestion)

src/routes/MapExplorer/MapDrawer/MapDrawer.js
```diff
@@ -43,121 +52,165 @@ function MapDrawer (props) {
...
+            routeParams={routeParams}
+            post={p}
+            styleName='contentCard'
+            constrained
+            expanded={false}
+            key={p.id}
+          />)}
+        </div>
+
+        <ScrollListener onBottom={() => fetchPostsForDrawer(numFetchedPosts, false)} elementId='mapDrawerWrapper' />
```
thomasgwatson, line 157: Does this one also fetch groups then now?

thomasgwatson: or wait, I think you send you just request all public groups

thomasgwatson (approved):
lgtm

src/routes/MapExplorer/MapDrawer/MapDrawer.js
tibetsprague: fetchGroups is its own query https://github.com/Hylozoic/hylo-evo/blob/616-map-load-performance/src/routes/MapExplorer/MapExplorer.store.js

yeah i just fetch all public groups in a bounding box for now

src/components/Widget/GroupsWidget/GroupsWidget.js
tibetsprague: Yeah thats what I decided to use in the map drawer for now (at Aaron’s suggestion)

Issue #974, opened 2021-10-26
When people click outside (or close) the create post modal ask them if they really want to discard their post


tibetsprague:
Right now when someone enters information into the new post dialog, if they accidentally click outside the dialog (or click the close button) it will close and they will lose their information. We should detect whether any of the fields in the form has data in it and use a confirm() popup to ask them if they really want to cancel post creation, letting them know they will lose all the data they've input so far. 

This will involve adding a document level click event listener when the post creation modal is open and removing it when it closes. See src/components/PostEditorModal/PostEditorModal.js and src/components/PostEditor/PostEditor.js

Bug report in Intercom: https://app.intercom.com/a/apps/wwelodje/inbox/inbox/all/conversations/27244381152

tibetsprague:
Text of the confirm dialog can say something like: "Are you sure you want to exit? You will lose all the information you have entered"

PR #1097, opened 2022-02-07
Confirm close on create post

antonetteortiz:
[#974 Confirm close on Create Modal ](https://github.com/Hylozoic/hylo-evo/issues/974)

Description
--
If user makes any changes to the Create Modal and clicks outside of the window or closes the modal, a confirmation prompt will appear. 

How to Test:
---
- In the create discussion modal, make changes to any field.
- Click outside of the window and/or attempt to close the modal.
- Confirmation window will popup, prompting to either return to the create modal or discard post.

### DIFF
--- a/src/components/CreateModal/CreateModal.connector.js
+++ b/src/components/CreateModal/CreateModal.connector.js
@@ -15,15 +15,19 @@ export const mapDispatchToProps = (dispatch, props) => {
   const closeUrl = postId
     ? postUrl(postId, urlParams, querystringParams)
     : addQuerystringToPath(baseUrl(urlParams), querystringParams)
-  const closeModal = () => dispatch(push(closeUrl))
+
+  const closeModal = (isDirty) => {
+    if (isDirty) {
+      if (window.confirm('Are you sure you want to exit? You will lose all the information you have entered?')) {
+        dispatch(push(closeUrl))
+      }
+    } else {
+      dispatch(push(closeUrl))
+    }
+  }
 
   return {
-    closeModal,
-    confirmLeaveCreateModal: () => {
-      if (window.confirm(`Are you sure you want to exit? You will lose all the information you have entered`)) {
-        closeModal()
-      }
-    }
+    closeModal
   }
 }
 
--- a/src/components/CreateModal/CreateModal.connector.js
+++ b/src/components/CreateModal/CreateModal.connector.js
@@ -15,9 +15,15 @@ export const mapDispatchToProps = (dispatch, props) => {
   const closeUrl = postId
     ? postUrl(postId, urlParams, querystringParams)
     : addQuerystringToPath(baseUrl(urlParams), querystringParams)
+  const closeModal = () => dispatch(push(closeUrl))
 
   return {
-    closeModal: () => dispatch(push(closeUrl))
+    closeModal,
+    confirmLeaveCreateModal: () => {
+      if (window.confirm(`Are you sure you want to exit? You will lose all the information you have entered`)) {
+        closeModal()
+      }
+    }
   }
 }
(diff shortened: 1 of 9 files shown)


src/components/CreateModal/CreateModal.connector.js
```diff
@@ -15,9 +15,15 @@ export const mapDispatchToProps = (dispatch, props) => {
   const closeUrl = postId
     ? postUrl(postId, urlParams, querystringParams)
     : addQuerystringToPath(baseUrl(urlParams), querystringParams)
+  const closeModal = () => dispatch(push(closeUrl))
 
   return {
-    closeModal: () => dispatch(push(closeUrl))
+    closeModal,
+    confirmLeaveCreateModal: () => {
```
tibetsprague, line 22: this works great, but could be slightly less code and cleaner by just passing an isDirty parameter into closeModal, then you wouldn't need two functions

src/components/CreateModal/CreateModal.js
```diff
@@ -8,30 +8,62 @@ import PostEditor from 'components/PostEditor'
...
-        <Switch>
-          <Route path={match.path + `/post`} children={({ match, location }) =>
-            <PostEditor match={match} location={location} onClose={closeModal} />} />
-          <Route path={match.path + `/group`} children={({ match, location }) =>
-            <CreateGroup match={match} location={location} onClose={closeModal} />} />
-          <Route>
-            <CreateModalChooser match={match} location={location} />
-          </Route>
-        </Switch>
+  const checkBeforeClose = () => {
```
tibetsprague, line 16: if you pass in isDirty to closeModal then this conditional moves to closeModal

src/components/PostEditor/PostEditor.js
```diff
@@ -106,7 +106,8 @@ export default class PostEditor extends React.Component {
       toggleAnnouncementModal: false,
       showPostTypeMenu: false,
       titleLengthError: false,
-      dateError: false
+      dateError: false,
+      isDirty // prop tracking if form fields were touched from CreateModal
```
tibetsprague, line 110: it doesnt look like you are actually using this, why do you need to pass isDirty down to the PostEditor?

src/components/PostEditor/PostEditor.js
```diff
@@ -118,6 +119,7 @@ export default class PostEditor extends React.Component {
     this.editor = React.createRef()
     this.groupsSelector = React.createRef()
     this.topicSelector = React.createRef()
+    this.setIsDirty = props.setIsDirty
```
tibetsprague, line 122: you dont need to do this, you can just call this.props.setIsDirty down below

src/components/PostEditor/PostEditor.js
```diff
@@ -278,6 +284,8 @@ export default class PostEditor extends React.Component {
       this.setState({
         detailsTopics: topicNames.map(tn => ({ label: tn, name: tn, id: tn }))
       })
+      // TODO (Annie): figure out why topic changes don't update isDirty
```
tibetsprague, line 287: hmm, strange, we can look at this together

tibetsprague (changes requested):
whoo hoo, looking good! just a couple little suggestions/changes requested, and we can look at the topics thing together

tibetsprague (commented):
oh also we will want to add/tweak our tests to cover the changes, so good opportunity to look at our test code together too

antonetteortiz:
thanks Tibet, I will work on the change request and connect with you later today to look at topic changes.

tibetsprague (approved):
i see you went ahead and cleaned up a lot of code, thanks :)

PR #1102, opened 2022-02-10
EVO-1055 allow correct dropdown menu items to display for all comments

Sofacosta:
I commented out some code on MapExplorer.store.test.js because it wasn't been used and it was failing some tests.


tibetsprague (approved):
great!

PR #1112, opened 2022-02-20
Fix PostEditor when not used in CreateModal context

lorenjohnson:
#974 breaks the `PostEditor` when used in other contexts beyond `CreateModal` as it relies upon `this.props.setIsDirty` existing. `PostEditorModal` which is used for editing and creating Posts when not through the `CreateModal` currently doesn't provide this. 

This was in my line-of-sight as I've been working away on both `HyloEditor` and `PostEditor` for the text formatting cluster of things. I went ahead and made a follow-up PR applying the confirmation logic to the `PostEditorModal` such that it is universally applied. Note that this also closes #1097.

* Also this PR modifies `ChangeModal` to use the `PostEditorModal` vs its own instance of `PostEditor`.


src/components/PostEditorModal/PostEditorModal.test.js
```diff
@@ -1,8 +1,47 @@
...
+//   const { getByPlaceholderText } = render(<PostEditorModal match={{}} location={{ search: '' }} />, null, providersWithStore)
+//   expect(await getByPlaceholderText('Type group name...')).not.toBe(null)
+// })
+
+// afterEach(() => {
+//   graphqlMockServer.resetHandlers()
+//   providersWithStore = null
+// })
+
+it.skip('needs tests that rely upon React Testing Library', () => {})
```
lorenjohnson, line 47: Note: The above can be uncommented and will pass once the related React Testing Library helpers are brought-in from the Group Explorer.

tibetsprague (approved):
ah dang, thanks for catching

PR #1113, opened 2022-02-21
When going back to map view keep center where it was

tibetsprague:
Fixes https://github.com/Hylozoic/hylo-evo/issues/1099


thomasgwatson (commented):
lgtm

PR #1115, opened 2022-02-22
Fix post editor modal

tibetsprague:
- Couldn't click outside to close
- Remove double close button
- Loading Hylo on an edit URL for a request would break

Removes PostEditorModal and moves all functionality to CreateModal


src/components/PostEditor/PostEditor.js
```diff
@@ -147,7 +147,8 @@ export default class PostEditor extends React.Component {
 
   componentDidUpdate (prevProps) {
     const linkPreview = this.props.linkPreview
-    if (get('post.id', this.props) !== get('post.id', prevProps)) {
+    if (get('post.id', this.props) !== get('post.id', prevProps) ||
+        get('post.details', this.props) !== get('post.details', prevProps)) {
```
lorenjohnson, line 151: Makes sense to me, yet this can be a touchy area given that if the details are changing while the component is mounted this would  possibly overwrite an in-progress edit. What came-up?

src/graphql/fragments/groupQueryFragment.js
```diff
@@ -159,6 +159,7 @@ export default () => {
       items {
         id
         title
+        type
```
lorenjohnson, line 162: Just curious why this came-up too?

lorenjohnson (approved):
Curious about a couple things, but looks great!

src/components/PostEditor/PostEditor.js
tibetsprague: hmm, theres a weird thing where if you load a request directly it breaks because we load the openOffersAndRequests for the group which loads some fields for the post but not all the details for that post. then it wasnt updating the post info when the full post got loaded

src/graphql/fragments/groupQueryFragment.js
tibetsprague: same issue, not having the type was breaking the page completely. we could just load all the post stuff here, but i was trying to reduce the load time of the explore page where this is used by only loading what is needed for the post cards there.

Issue #833, opened 2021-03-17
Invite expired page is broken/empty


tibetsprague:
Probably a routing related issue

- [ ] Going to an invalid invite link goes to infinite spinner, should show an error message. There is an error here that is not being caught/handled
- [ ] Also check an invite link that was sent to someone by email and then expired

What should it show? An invite expired error page with the (Jolly/sad?) Axolotl, some work to do here

thomasgwatson:
How do we easily replicate invalid invite links?
- Can you invite someone, and then revoke the invite? Does that invalidate the invite?
@tibetsprague @lorenjohnson ?

tibetsprague:
Also you can just just change a letter in the link :)

thomasgwatson:
Ahh yes excellent! Just chatted with Annie about a next ticket.

lorenjohnson:
FYI `checkInvitation` is not returning the correct result for expired invitations on the backend (for direct invites with `invitationToken` vs the open Group Invite URL) , addressing that then will look into the flow of things once there is an expired invite both in the Auth and Non-Auth contexts.

PR #1130, opened 2022-03-03
Refactor registration to work with users created by API

tibetsprague:
Remaining work:

- [x] When following a join group link the group will be joined after login or signup, this however not persist page reloads while login and signup state obviously will... This wouldn't be terribly hard to remedy using a cookie for `returnToURL` or even possibly adding a `user.settings.returnToOnSignup`? *FWIW, this has always been the case, but will be an issue with email verification links*
- [x] Merge current `dev`
- [x] QA focused on signup, join group via signup,  email verification as a currently verified user, create modal in different contexts
- [x] Fix CircleCI test runs
- [x] Create Group Modal duplicate/unavailable slug errors are appearing late or not at all (a little flakey sometimes, but nothing new. Will create ticket for needed improvements)
- [x] Create Group Modal close and back button both go back to Welcome Tour final screen, which kinda makes sense but the close button shouldn't do that 
- [x] Gather name, not just email, on Google Signup
- [x] Handle Expired Group Invitations by figuring-out how to honor the redirect to the proxied page or passing an error through the SignUp page. Also consider signed-in case: https://github.com/Hylozoic/hylo-evo/issues/833
- [x] Confirm that legacy `/m`, `/u`, `/tag` URLs which are used in old posts are already transformed as expected in `HyloShared.TextHelpers` or `ClickCatcher` etc? otherwise add back these legacy route redirects.
- [x] Need option for cancelling from Registration page (Name/Password)
- [x] Merge-up `dev` branch and test
- [x] Fix Create Post from Map goes to Map Post Detail issue
- [ ] #1151 Re-sent link on Validation Email
- [ ] Add CHANGELOG entry
----

Draft CHANGELOG entry:

- Adds `store/selectors/getSignupState` as primary source of getting authentication/authorization state from `currentUser`, see notes in that file for more details
- Adds `SignupRouter` to handle implicit signup state routing
- Moves  `routes/RootRouter` to separate `react-router` setup from actual routing, `RootRouter` is where all routing begins and authorization is checked. This all used to happen in `router/index`  / `AuthRoute`
- All non-authorized routing happens in `NonAuthLayoutRouter` (was `NonAuthLayout`)
- All public routing and layout happens in the new `PublicLayoutRouter` 
- Significantly routing refactors in `PrimaryLayout` renamed to `AuthLayoutRouter`
- Moves the site tour to it's own component `SiteTour`
- `WelcomeWizardModal` becomes `WelcomeWizardRouter` and now handles it's own routes
- A handful of other component renames under `src/routes`, most converted from connected class components to functional components using hooks
- React Testing Library (RTL) setup simplified and notated
- Added a Mock Service Worker setup to create a standard `mockGraphqlServer` for use in tests
- New tests added, in particular for components converted from class to functional. These all use RTL and some also use `mockGraphqlServer` so serve as good examples for subsequent testing
- Renames `errorMiddleware` to `rollbarMiddleware` as that's what it is and was otherwise confusing in terms of other error handling (i.e. graphql response errors)
- An annoying amount of linting updates which make life nice in the code, but make this PR a bit of a headache and I'm sorry for that
- Refactor of `CreateModal` and a simplification of its routing
- Other things related to the above...

Follow-ups for another time:

- Graphql request errors arrive at the frontend in potentially 2 ways:
  - User/auth mutations assume: `payload.data.<operationName>.error`. These are currently handled in the components which use them.
  - `payload.errors`: `graphqlMiddleware` throws when there is a value for `payload.errors` taking the first of the `payload.errors` array and re-throwing it. To normalize I'd be tempted to mung `payload.data` adding an `payload.<operationName>.error` add that first error, making the results throughout work like the backend for the auth/signup related resolvers.
  - Both of these errors can exists in the case of the user/auth mutations...
- Sometime it'd be nice to also convert `FullPageModal` to expect routes and use those routes in the related components which would be renamed `GroupSettingsRouter` and `UserSettingsRouter` accordingly
- Can probably entirely eliminate `connected-react-router`, we are not relying on its features and it's adding some extra noise around our routing setup

-----

This means we now have partially created users that are inactive but logged in as soon as a person enters their email to register

Fixes https://github.com/Hylozoic/hylo-evo/issues/1124


src/routes/AuthLayoutRouter/AuthLayoutRouter.js
```diff
@@ -413,105 +252,115 @@ export default class PrimaryLayout extends Component {
...
+  { path: `/:context(groups)/:groupSlug/:view(members)/:personId/${POST_DETAIL_MATCH}`, component: PostDetail },
+  { path: `/:context(groups)/:groupSlug/:view(explore|map|groups)/${GROUP_DETAIL_MATCH}`, component: GroupDetail },
+  { path: `/:context(groups)/:groupSlug/:view(topics)/:topicName/${POST_DETAIL_MATCH}`, component: PostDetail },
+  { path: `/:context(groups)/:groupSlug/${POST_DETAIL_MATCH}`, component: PostDetail },
+  { path: `/:context(groups)/:groupSlug/${GROUP_DETAIL_MATCH}`, component: GroupDetail },
+  { path: `/:view(members)/:personId/${POST_DETAIL_MATCH}`, component: PostDetail }
+]
+
+// Redirects for legacy and NonAuth routes
+const redirectRoutes = [
```
tibetsprague, line 340: i wonder if/when we could get rid of a lot of these redirects?

src/routes/AuthLayoutRouter/components/Drawer/Drawer.js
```diff
@@ -0,0 +1,131 @@
+import React from 'react'
+import { Link, useLocation, useHistory } from 'react-router-dom'
+import { get } from 'lodash/fp'
```
tibetsprague, line 3: would love to alphabetize routes as we go. not a big deal of course

src/routes/JoinGroup/JoinGroup.js
```diff
@@ -1,65 +1,74 @@
...
           const useInvitationResult = await dispatch(useInvitation(invitationTokenAndCode))
           const newMembership = useInvitationResult?.payload?.getData()?.membership
-          setGroupSlug(newMembership?.group?.slug)
+          const groupSlug = newMembership?.group?.slug
+
+          if (groupSlug) {
+            /*
+              `AuthLayoutRouter` will already try and fetch this group due to the
+              `/groups/:groupSlug/join/<token>` route matching `:groupSlug` before the
+              group has been joined (unauthorized), this could be fixed and this extra
```
tibetsprague, line 44: file a ticket for this?

tibetsprague: join group is definitely slower than it needs to be because it loads the entire me query when it doesnt need to, seems like this is fixed here which is great!

src/routes/RootRouter/RootRouter.js
```diff
@@ -0,0 +1,64 @@
...
+  if (isAuthorized) {
+    return (
+      <Route
+        path={[
+          // The routes matching here serves to provide `match.params` to `AuthLayoutRouter`.
+          // It likely better to move away from this, pushing the related routes and features
+          // into their own components below `AuthLayoutRouter` receiving the `match.params`
+          // also already happening from the routes found there. Search for the following
+          // comment to see the places where the `match.params` supplied here are currently
+          // replied-upon:
```
tibetsprague, line 43: relied-upon

tibetsprague (commented):
honestly nothing really jumps out at me, time to start testing!

src/routes/AuthLayoutRouter/AuthLayoutRouter.js
lorenjohnson: Yes please! I see no real reason not to. Worst case we'll find a notification email that still has an old URL in it, but that's unlikely at this point. I think any notification emails or bookmarks folks had from earlier days would be gone by now.

lorenjohnson: I just did it and along with it re-integrated the `DrawerRouter` routes into the layout. It really isn't the best candidate for splitting things out anyway. Things are pretty tidy now.

src/routes/JoinGroup/JoinGroup.js
lorenjohnson: Fixed this.

src/routes/AuthLayoutRouter/AuthLayoutRouter.js
lorenjohnson: All gone!

src/routes/RootRouter/RootRouter.js
```diff
@@ -33,23 +33,7 @@ export default function RootRouter () {
...
-          // The routes matching here serves to provide `match.params` to `AuthLayoutRouter`.
-          // It likely better to move away from this, pushing the related routes and features
-          // into their own components below `AuthLayoutRouter` receiving the `match.params`
-          // also already happening from the routes found there. Search for the following
-          // comment to see the places where the `match.params` supplied here are currently
-          // replied-upon:
-          //
-          // "NOTE: Relies on incoming `match.params` (currently provided by `RootRouter`)"
-          //
-          '/:context(groups)/:groupSlug/:view(events|groups|map|members|projects|settings|stream|topics)?',
```
tibetsprague: nice, what did you do?

tibetsprague (commented):
i can do more testing, but maybe ready to go??

src/routes/RootRouter/RootRouter.js
lorenjohnson: Well basically just moved it back into `AuthLayoutRouter` to approx. what you were doing before, but with nailed it down more. There were already a couple other `matchPaths` going on there still which I consolidated all into one place and put all (one exception) the flags dependent on this preliminary route matching (`pathMatchParams`) right next to it in the component. The idea here is that as we move forward we should be reducing and eventually mostly eliminating this pre-match in deference to pushing things into their own child router components... Something like that:

https://github.com/Hylozoic/hylo-evo/blob/1987e1c15e5cbe6e5fc96f53e1916ae81c5fac7a/src/routes/AuthLayoutRouter/AuthLayoutRouter.js#L66-L77

lorenjohnson:
> i can do more testing, but maybe ready to go??

Yep! I finished my own testing / fixing yesterday and very happy with where things are at at this point. I say we merge it down, but do note the remaining small fix needed in `CreateGroupModal` and adding the `CHANGELOG` entry which are the things on the "tasks lists" still open here on the PR. I have those items on my personal list over here so I don't need them track here, so just FYI. I will get to them before release release. It would be relieving to have this merged to `dev` so we can get on with merging it back up to WIP PRs and handle any conflicts and such there before things drift too far apart.

src/routes/AuthLayoutRouter/AuthLayoutRouter.js
```diff
@@ -176,27 +175,16 @@ export default function AuthLayoutRouter (props) {
         <RedirectRoute exact path='/:context(groups)/:groupSlug' to={`/groups/${currentGroupSlug}/explore`} />
       )}
 
-      <Route path='/:context(groups)/:groupSlug' render={routeProps => <GroupWelcomeModal {...routeProps} />} />
+      {!hyloAppLayout && (
```
lorenjohnson, line 178: @tibetsprague here is where the `hyloApp` (previously `mobileSettings`) the layout flags begin to diverge from the alias we're currently using of `withoutNav`. It's your call on whether you want to use a custom layout flag (`cafeLayout`?) or simply start using the general `withoutNav` as the actual querystring flag (e.g. `?layoutFlags[]=withoutNav|cafeLayout`). I think the more custom flag is maybe best. Regardless, whatever you do there you'll want to keep `hyloAppLayout` added to the conditionals where they currently are also. Hope that makes sense. Reach-out if you want clarification.

lorenjohnson: The only thing you need to start using an additional layout flag is add it's name to the whitelist in `LayoutFlagsContext` and just add it to the querystring param `layoutFlags`, noting that the word `Layout` is appended by the context so the flag itself shouldn't have that in the name.

PR #1131, opened 2022-03-03
(Addendum) Add text formatting && Align text handling across node, Web, and App

lorenjohnson:
This addendum was needed due to an issue with running the React Testing Library test once I merged into this branch. The root cause of that ended-up being related to a `jest` timers configuration change on the move up to jest 27 from 26. It took a long time to get to the bottom of this one. Anyway, it's a welcome change as timers have confusing and a debatably a little broken in Jest for awhile until this release. 

* Result of merge of current `dev` into this branch (so this includes the Group Explore changes.  I apologize that you'll be filtering through some of that code again...)

There is some extra in here, as there was in the parent branch, and mostly a result of getting to the bottom of what was going wrong with the `msw/graphql` + `@testing-library/react` setup in this branch including:

* Breaking-out our StandardJS config to it's equivalent custom ESLint configuration to allow addition of custom rules (helpers really) for Jest and React Testing Library. This configuration will not currently be used by anything other than the ESLint VSCode extension and should does not change our current linting using the Standard binary, our `precommit` hook, WebPack build, or `yarn lint`. However, my recommendation is that after this merge we all use the ESLint extension in VSCode and turn off the Standard extension (they're mutually exclusive). This will provide the same linting and turn-on the help in `*.test.js` files. It will also have the effect of applying Standard v16 rules in the editor while for the moment maintaining the more tolerant Standard v12 rules on `precommit` (currently about 2000 rule breaks with this ESLint Standard v16 config!).  \*

* Added the root workspace `.vscode` directory to the project which includes the essential extension recommendations and configurations.

* Slightly simplify `GroupSearch.test.js` as the reference example of applying ReactTestingLibrary and using the MockServiceWorker (`msw`) for graphql responses. Also updated and enabled the slightly simpler `CreateModal.test.js`.

* Related dependency upgrades

* Also, including in the original branch which you already reviewed @tibetsprague I forgot to mention that there are a few things here which remove all the noise from the tests output. 

_\* The idea here is to create a path for gradually moving-up to soften the move up to the current StandardJS version. All of the new rules above v12 are things I personally am a fan of, and have been doing anyway. The custom `eslint` configuration enables us to move-up more gradually by turning some rules that would be standard errors to warnings. Generally it's nice to not have to deal with `eslint` config by using the `standard` bin, but I think at the point we're at now it makes sense to take a bit more control (e.g. for things like these testing `eslint` plugins) while otherwise maintaining the standard rules._


src/components/CreateModal/CreateModal.test.js
```diff
@@ -1,47 +1,29 @@
...
+  )
 
-// beforeEach(() => {
-//   const session = orm.mutableSession(orm.getEmptyState())
-//   session.Me.create({ id: '1' })
-//   const initialState = { orm: session.state }
-//   const store = generateStore(history, initialState)
-//   providersWithStore = AllTheProviders(store)
-// })
+  expect(getByText('What would you like to create?')).toBeTruthy()
```
thomasgwatson, line 24: what do you think about this being the functionally the same but slightly more descriptive `.toBeInTheDocument`?

lorenjohnson: I like that better, will change. Was referencing docs which will sometimes do a `toBeTruthy` because they're not relying on the extended matchers.

thomasgwatson:
Couldn'y quite figure out what changes you made to src/routes/GroupExplorer/GroupSearch/GroupSearch.test.js ? Did they get committed?

tibetsprague (approved):

lorenjohnson:
> Couldn'y quite figure out what changes you made to src/routes/GroupExplorer/GroupSearch/GroupSearch.test.js ? Did they get committed?

Yup. Sorry to say that the diff isn't gonna show evidently due to the `dev` -> [upgrade-draft-js-plugins](https://github.com/Hylozoic/hylo-evo/tree/upgrade-draft-js-plugins) <- [upgrade-draft-js-plugins-addendum](https://github.com/Hylozoic/hylo-evo/tree/upgrade-draft-js-plugins-addendum) merge path. Tried to generate a good diff here on GitHub to link back to you but I think because I made the changes as part of a merge commit it's still showing as all new changes on the diff with `dev` too 😔 So, I guess you're gonna have to eyeball it.

Here is where there were things were noting --

```
test('GroupSearch integration test', async () => {
  const { queryByText, getByRole } = render(<GroupSearch />, container, providersWithStore)

  expect(queryByText('Group Search')).toBeInTheDocument()
  await waitFor(() => expect(queryByText('Test Group Title')).toBeInTheDocument())

  fireEvent.change(getByRole('textbox'), {
    target: { value: 'different group' }
  })
  await waitFor(() => expect(queryByText('Search input results')).toBeInTheDocument())
})
```

Revised:

```
test('GroupSearch integration test', async () => {
  const { findByText, getByRole } = render(<GroupSearch />, null, providersWithStore)

  expect(await findByText('Test Group Title')).toBeInTheDocument()

  fireEvent.change(getByRole('textbox'), {
    target: { value: 'different group' }
  })

  expect(await findByText('Search input results')).toBeInTheDocument()
})
```

Though still supported and sometimes useful `waitFor` is turning-out to be an anti-pattern in RTL with awaiting `findBy...` methods which return promises to be preferred. Also the `container` DOM node didn't end-up being necessary.

PR #1138, opened 2022-03-11
Feature/evo 1046 click map to create post

Sofacosta:


src/components/CreateModal/CreateModalChooser.js
```diff
@@ -11,14 +11,16 @@ const postTypes = Object.keys(POST_TYPES)
 export default class CreateModalChooser extends Component {
   render () {
     const { location } = this.props
+    const pathname = location.pathname.slice(-1) === '/' ? location.pathname : `${location.pathname}/`
+    const params = location.search ? `&${location.search.substring(1)}` : ''
 
     return <div styleName='chooser'>
-      <h1>What would you like to create?</h1>
+      <h1>{params.includes('lat') && params.includes('lng') ? 'New Post at this location: ' : ''}What would you like to create?</h1>
```
lorenjohnson, line 18: For clarity maybe make this `const hasLocation = params.includes(`lat`) && ...` above the `return` and then use that in the render...

src/components/PostEditor/PostEditor.js
```diff
@@ -545,12 +546,17 @@ export default class PostEditor extends React.Component {
       'project'
     ].includes(type)
     const canHaveTimes = type !== 'discussion'
-    // Center location autocomplete either on post's current location, or current group's location, or current user's location
+    const params = windowLocation && windowLocation.search ? Object.fromEntries(new URLSearchParams(windowLocation.search)) : null
+    let lat, lng
+    if (params) {
+      lat = params.lat
+      lng = params.lng
+    }
```
lorenjohnson, line 554: I think it'd be better if `PostEditor` didn't have to know about the location for this but instead received either `location` or `locationObject` as a prop (or `post.location`) coming from `CreateModal` which handled grabbing the querystring/search params then passed them.  Similarly I think you can handle `currentLocation` by constructing the map lat/lng point into a `locationObject` you pass in as that prop.

src/components/PostEditor/PostEditor.js
```diff
@@ -696,7 +702,7 @@ export default class PostEditor extends React.Component {
               <LocationInput
                 saveLocationToDB
                 locationObject={curLocation}
-                location={location}
+                location={location || `${lat}, ${lng}`}
```
lorenjohnson, line 705: If you setup the value for location in the prop you're sending in then I think this extra `||` should be able to be eliminated.

src/routes/MapExplorer/MapExplorer.js
```diff
@@ -62,6 +62,8 @@ export class UnwrappedMapExplorer extends React.Component {
       groupIconLayer: null,
       hideDrawer: props.hideDrawer,
       hoveredObject: null,
+      canCreatePost: false,
```
lorenjohnson, line 65: Is this maybe `creatingPost` vs `canCreatePost` ? I may be missing something but I think that naming signifies better how this state value is being used.

lorenjohnson (changes requested):
Looks at the comments and respond back here or just hmu if you any of that is unclear or you want help. Generally looks good!

lorenjohnson (approved):
Yay, looks good! 🎉

Issue #1137, opened 2022-03-11
group skills not working for new group members


cpolitano:
When entering a group that has suggested Skills and Interests, the skills pills are not interactive.

Expected behavior: clicking on a pill should highlight the pill and add it to the member's profile

![Screen Shot 2022-03-09 at 1 06 16 PM](https://user-images.githubusercontent.com/5813534/157795206-852a6117-a1f6-4785-9498-7ca5c11ff730.png)

PR #1140, opened 2022-03-14
Fix adding recommended skills when joining a new group

tibetsprague:
Fixes https://github.com/Hylozoic/hylo-evo/issues/1137


thomasgwatson (approved):
lgtm

PR #1144, opened 2022-03-17
click on map additional changes

Sofacosta:
https://github.com/Hylozoic/hylo-evo/pull/1138


src/store/selectors/getQueryParamsObjectFromString.js
```diff
@@ -0,0 +1,10 @@
+const getQueryParamsObjectFromString = (queryParamString) => {
+  let obj = {}
+  const params = new URLSearchParams(queryParamString)
+  for (const [key, value] of params) {
+    obj[key] = value
+  }
+  return obj
+}
+
+export default getQueryParamsObjectFromString
```
Sofacosta, line 10: Similar functions live in selectors however it feels like more of a util since it doesn’t get anything from the redux store. Should this function live here or somewhere else?

src/routes/MapExplorer/MapExplorer.js
```diff
@@ -464,87 +469,89 @@ export class UnwrappedMapExplorer extends React.Component {
...
+        </div>
+        <button styleName={cx('toggleDrawerButton', { 'drawerOpen': !hideDrawer })} onClick={this.toggleDrawer}>
+          <Icon name='Hamburger' className={styles.openDrawer} />
+          <Icon name='Ex' className={styles.closeDrawer} />
+        </button>
+        {!hideDrawer && (
+          <MapDrawer
+            context={context}
+            currentUser={currentUser}
+            fetchPostsForDrawer={fetchPostsForDrawer}
```
Sofacosta, line 494: White space changes

src/components/PostEditor/PostEditor.js
```diff
@@ -544,17 +542,21 @@ export default class PostEditor extends React.Component {
 
     const hasStripeAccount = get('hasStripeAccount', currentUser)
     const hasLocation = [
+      'discussion',
```
tibetsprague, line 545: well we never officially decided this is what we want to do, but i like it, i will double check with the rest of the team

src/components/PostEditor/PostEditor.js
```diff
@@ -544,17 +542,21 @@ export default class PostEditor extends React.Component {
...
       'event',
       'offer',
       'request',
       'resource',
       'project'
     ].includes(type)
     const canHaveTimes = type !== 'discussion'
```
tibetsprague, line 552: wouldnt you want to remove this?

tibetsprague:
This is looking really great! There are a couple things left i see, so close...
- [ ] It seems that location saving broke completely so new posts i create from the map are no longer saving with their location. Not sure what happened here. 
- [ ] The + button isn't quite in the right place when the drawer is open, it should be further to the right, aligned with the X button at the top that closes the drawer. 
- [ ] There should be a tooltip on the plus button that says "Add post to the map". You can see how we use <ReactTooltip> elsewhere in the codebase
- [ ] And sorry to be a stickler, but there's a couple more places where it would be good to keep the query params (and this is broken on production too) 1) when you close the post creation dialog 2) When you open a post from the drawer. 
- [ ] We need to add Discussions to the Features filter button on the bottom left (should be at the top)

src/components/PostEditor/PostEditor.js
```diff
@@ -565,13 +565,13 @@ export default class PostEditor extends React.Component {
         <div styleName='header'>
           <div styleName='initial'>
             <div>
-              <Button {...this.postTypeButtonProps(type)} />
+              <Button noDefaultStyles {...this.postTypeButtonProps(type)} />
```
tibetsprague, line 568: im confused about what changed here that requires this

tibetsprague (approved):
still a bit confused about why the button style needs to be turned off suddenly, but otherwise this looks good!

tibetsprague:
dang, so close, but one other lingering bug! for some reason as i zoom around sometimes the location for the post gets replaced by undefined, undefined..?? Then as i zoom around more it goes back to the correct display of the lat,long. So weird?? probably something to do with a "location" prop name being reused or something? last thing to figure out...
Screen Shot 2022-03-31 at 10 17 46 PM

src/components/PostEditor/PostEditor.js
Sofacosta: I am also confused why this requires the default styles to be off.

tibetsprague (approved):

PR #1145, opened 2022-03-21
Evo 947 farm collaboration

### LINKED ISSUE #947: Add `farm-collaborate` widget
A widget that shows what forms of collaboration farmers are open to

thomasgwatson:
closes #947


src/components/Icon/Icon.js
```diff
@@ -14,7 +15,7 @@ export default function Icon ({
   const iconClassName = `icon-${name}`
 
   return <span
-    styleName={cx('icon', { green })}
+    styleName={cx('icon', { green, blue })}
```
tibetsprague, line 18: how can an icon be both green and blue?

src/store/selectors/getFarmData.js
```diff
@@ -0,0 +1,6 @@
+import { FARM_ONBOARDING } from 'util/constants'
+
+export default function getFarmData (group) {
```
tibetsprague, line 3: this is not exactly a selector... but ok

src/store/selectors/isGroupRoute.js
```diff
@@ -10,7 +10,7 @@ const isGroupRoute = createSelector(
 export default isGroupRoute
 
 export function getSlugFromLocation (state, props) {
-  return getRouteParam('groupSlug', state, props, false) || tryLocation(props) || props.groupSlug
+  return getRouteParam('groupSlug', state, props, false) || getRouteParam('detailGroupSlug', state, props, false) || tryLocation(props) || props.groupSlug
```
tibetsprague, line 13: this makes me nervous, is a detailGroup the same as a group route? are we sure this doesnt cause issues in some other cases of checking isGroupRoute?

src/store/selectors/getFarmData.js
thomasgwatson: What's your name for a function that looks up data from the store?

src/components/Icon/Icon.js
thomasgwatson: Those are props. But if someone passed in `green = true` and `blue = true`, then it would apply both classes but that would be obviously wrong for them to do that.

src/store/selectors/getFarmData.js
tibetsprague: to me this is a utility function on the group model, so it could go in Group.js. A selector is a specific type of memoized function. And in react/redux code it usually specifically means one that pulls data from the redux state. but I also am ok leaving this it here

src/store/selectors/isGroupRoute.js
tibetsprague: getGroupForDetail seems like a different case. we should really comment more on these things

src/components/Icon/Icon.js
tibetsprague: ah yes right

Issue #960, opened 2021-10-21
Can't create Event posts on small screens


tibetsprague:
Reported that people couldn't scroll down to see the full post creation dialog: https://app.intercom.com/a/apps/wwelodje/inbox/inbox/all/conversations/27244385048

tibetsprague:
Happening on windows chrome
What is happening is that the time part of the timeframe popup is hidden below her screen and she doesnt know how to scroll using a scroll wheel, and so when she goes to click on the scroll bar the timeframe popup disappears. 
One solution would be to have it pop up above the button so it definitely fully appears on the screen

PR #1147, opened 2022-03-22
Can't create Event posts on small screens #960

antonetteortiz:
"Happening on windows chrome
What is happening is that the time part of the timeframe popup is hidden below her screen and she doesnt know how to scroll using a scroll wheel, and so when she goes to click on the scroll bar the timeframe popup disappears.
One solution would be to have it pop up above the button so it definitely fully appears on the screen"


tibetsprague (commented):
hmm, i dont understand how this diff fixes the issue? what i remember from our conversation was that we were going to detect if some or all of the date picker would appear below the viewport and if so have it appear *above* the date picker button instead of below, so you could always be sure to see the whole datepicker. i cant see how this code would do that. what does this do?

thomasgwatson:
@antonetteortiz can you pop some screenshots of what the changes look like for where the date picker is now positioned, and where the time picker ends up?

And it looks like you decided you didn't need to edit the `datepicker.scss` file that we found at `src/css/datepicker.scss`?

tibetsprague (approved):

PR #1171, opened 2022-04-07
1135 flag to hide extension data

### LINKED ISSUE #1135: Add flag to turn on/off showing farm data on a group's about page
Store in settings JSONB column for the group. Call the flag 'hide_extension_data'.
Add toggle to main Group Settings page to "Show {group type} data on about page"
Only show this if the group has a special group type e.g. Farm
Defaults to on - showing the data (if the flag is not set)
This flag determines whether to show special farm data on the group about page

thomasgwatson:
closes #1135 

image


tibetsprague (commented):
not 100% convinced about Data Settings as the name of a new tab. what else will go here? lets think about all the new settings and how to best arrange

thomasgwatson:
Was thinking the "location privacy" setting would also go here. Is there
another name you think fits?

PR #1173, opened 2022-04-07
Node 767 hide extension data

thomasgwatson:
Front-end work to hide extension data. Currently includes #1135 work


tibetsprague (commented):
looks good, pending one discussion about how to organize/describe the settings

Issue #1047, opened 2022-01-12
Allow for changing base layer of the map


tibetsprague:
The ability to change the base map layer being displayed by MapBox to Satellite view or satellite with streets

Issue #1048, opened 2022-01-12
Show indigenous territories as layer on the map


tibetsprague:
From [native-land.ca](https://www.figma.com/file/hITe76b9jN7BJuapxuUBuN/Rebirth?node-id=11%3A50)

We probably pull this data and cache locally once per day

“If you’re using this dataset for your project and you want to properly credit it, it’s not too hard. You can give credit to “Native Land Digital”, and link to our website. The dataset is publicly available and is not under any proprietary license; the work can be considered under a CC0 license (free to use in any way you like). We would, however, kindly ask that you consider the morality of how you use this dataset, be aware that it is constantly changing, and that it is not meant or reasonably to use for academic or legal purposes."

thomasgwatson:
https://www.mapbox.com/mts

tibetsprague:
add a link to native lands

PR #1177, opened 2022-04-15
Add base layer selector + native territories layer to the map

tibetsprague:
Fixes https://github.com/Hylozoic/hylo-evo/issues/1047 and https://github.com/Hylozoic/hylo-evo/issues/1048

Almost ready, just needs a little styling love. 
@thomasgwatson you will notice that i decided to go with using react-map-gl layers for now because of limitations in DeckGl's layer stuff, like not being able to handle overlapping labels, and not being able to gather info from all the labels being hovered over at once. We can keep discussing as we add more layers how to handle them


thomasgwatson (approved):
In general, we can do interactive layers in deck.gl (what its built for). The defacto responsibility for baselayers lies with mapbox-gl, so this is fine

PR #1178, opened 2022-04-22
Evo 1068 upload geo json text for group shape

Sofacosta:


src/components/Map/EditableMap/EditableMap.js
```diff
@@ -0,0 +1,117 @@
+import React, { useCallback, useEffect, useRef, useState } from 'react'
+import MapGL from 'react-map-gl'
+import { Editor, DrawPolygonMode, EditingMode } from 'react-map-gl-draw'
+import Icon from '../../Icon'
```
tibetsprague, line 4: we like to use the absolute path of 'components/Icon', just in case this component moves

src/components/Map/EditableMap/EditableMap.js
```diff
@@ -0,0 +1,117 @@
...
+  const { locationObject, polygon, savePolygon } = props
+  const centerAt = locationObject?.center
+  const showLocationOnMap = !polygon && centerAt
+  const [viewport, setViewport] = useState(null)
+  const [mode, setMode] = useState(null)
+  const [selectedFeatureIndex, setSelectedFeatureIndex] = useState(null)
+  const [isModeDrawing, setIsModeDrawing] = useState(false)
+  const editorRef = useRef(null)
+  const existingPolygon = polygon ? [{
+    geometry: polygon,
```
tibetsprague, line 20: this is the only variable that changes, so less code to have the ternary operator only here for this line

src/components/Map/EditableMap/EditableMap.js
```diff
@@ -0,0 +1,117 @@
...
+    properties: { },
+    type: 'Feature'
+  }] : [{
+    geometry: { },
+    properties: { },
+    type: 'Feature'
+  }]
+
+  useEffect(() => {
+    const defaultLocation = showLocationOnMap ? {
```
tibetsprague, line 30: shouldnt this location be based on the polygon if that exists first?

src/components/Map/EditableMap/EditableMap.js
```diff
@@ -0,0 +1,117 @@
...
+  )
+
+  return (
+    <MapGL
+      {...viewport}
+      width='100%'
+      height='100%'
+      mapOptions={{ logoPosition: 'bottom-right' }}
+      attributionControl={false}
+      mapStyle='mapbox://styles/mapbox/light-v9'
```
tibetsprague, line 98: lets discuss which base layer we want to use

tibetsprague (commented):
looks pretty good! lets walk through a demo together and talk about what is next, maybe tomorrow?

src/components/Map/EditableMap/EditableMap.js
Sofacosta: This is getting location based on pin/precise location, not polygon location. It centers the map on `defaultLocation` which defaults to Oakland if no `locationObject` exists. 

in the future it would be great to find the center of the polygon location using a bounding box

tibetsprague:
still TODO:
- [ ] when group shape is empty it is saying "null", should just be empty

src/routes/GroupSettings/GroupSettingsTab/GroupSettingsTab.js
```diff
@@ -119,6 +127,7 @@ export default class GroupSettingsTab extends Component {
     } = edits
 
     const { locationDisplayPrecision, showSuggestedSkills } = settings
+    const editableMapLocation = group?.locationObject || currentUser.locationObject
```
tibetsprague, line 130: what does this do if there's no location for the group or user? there should be a default location, and/or we could use the browsers get current location functionality

tibetsprague:
One bug im noticing is that after you have saved a polygon clicking the delete polygon button does not delete it. if you have any more time @Sofacosta that would be a nice final thing to fix

tibetsprague (approved):

PR #1182, opened 2022-04-27
1179 Message group moderators

tibetsprague:
Also included in this PR is the creation of a new Privacy & Access group settings tab and moving all privacy, visibility and access settings there. 

The hide extension data setting is now there and instead of saying extension says Hide {type} data like Hide Farm data.


thomasgwatson (approved):
lgtm

PR #1195, opened 2022-05-05
989 Group About Video

lorenjohnson:
Requires: https://github.com/Hylozoic/hylo-node/pull/804

### DIFF
--- a/src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
+++ b/src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import { TextHelpers } from 'hylo-shared'
 import useEnsureCurrentGroup from 'hooks/useEnsureCurrentGroup'
 import useEnsurePosts from 'hooks/useEnsurePosts'
 import Widget from 'components/Widget'
@@ -18,7 +19,16 @@ export default function FarmGroupDetailBody ({
   // TODO: hide widgets if they have no data, means loading all the data here?
   const widgets = [
     { settings: {}, isVisible: true, name: 'farm_at_a_glance' },
-    { settings: { title: mission, text: group.description || bio }, isVisible: true, name: 'mission' },
+    {
+      name: 'mission',
+      isVisible: true,
+      settings: {
+        title: mission,
+        embeddedVideoURI: group.aboutVideoUri,
+        text: !!group.description && bio,
+        richText: TextHelpers.markdown(group.description)
+      }
+    },
     { settings: {}, isVisible: true, name: 'farm_details' },
     { settings: {}, isVisible: true, name: 'opportunities_to_collaborate' },
     { settings: {}, isVisible: true, name: 'relevant_requests_offers' },
--- a/src/components/Widget/RichTextWidget/RichTextWidget.js
+++ b/src/components/Widget/RichTextWidget/RichTextWidget.js
@@ -1,22 +1,23 @@
-import PropTypes from 'prop-types'
-import React, { Component } from 'react'
-
+import React from 'react'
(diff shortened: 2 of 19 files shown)


src/components/Widget/RichTextWidget/RichTextWidget.js
```diff
@@ -1,22 +1,23 @@
...
-      <div styleName='rich-text'>
-        <h2>{settings.title || `Welcome to ${group.name}!`}</h2>
-        <p>{settings.text || `We're happy you're here! Please take a moment to explore this page to see what's alive in our group. Introduce yourself by clicking Create to post a Discussion sharing who you are and what brings you to our group. And don't forget to fill out your profile - so likeminded new friends can connect with you!`}</p>
-      </div>
-    )
-  }
+export default function RichTextWidget ({ group, settings }) {
+  return (
+    <div styleName='rich-text'>
+      <GroupAboutVideoEmbed uri={settings.embeddedVideoURI} />
```
tibetsprague, line 8: so i believe this is where we decided not to include the video anymore right? since this is on the Explore page and we would want a different welcome type video here (in the future)

src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
```diff
@@ -18,7 +19,16 @@ export default function FarmGroupDetailBody ({
...
   const widgets = [
     { settings: {}, isVisible: true, name: 'farm_at_a_glance' },
-    { settings: { title: mission, text: group.description || bio }, isVisible: true, name: 'mission' },
+    {
+      name: 'mission',
+      isVisible: true,
+      settings: {
+        title: mission,
+        embeddedVideoURI: group.aboutVideoUri,
+        text: !!group.description && bio,
```
tibetsprague, line 28: this is a little confusing, worth a comment i think saying that if richText is available use that. i am actually not totally sure why we need both fields, cant we just add markdown to the description if it exists and use the one text field?

lorenjohnson: See notes on Slack. This is due to the farm `bio` field. Both options are needed for now.

src/components/Widget/RichTextWidget/RichTextWidget.js
lorenjohnson: This is what appears on the group about for a farm.

lorenjohnson:
@tibetsprague here and your comment on node are correct. Neither PR yet includes the functionality to add this URL, simply spaced it and haven't added that yet. I will and flag you again. The other things you commented on though are correct and were the same things tripping me up in implementing it. It's confusing, but I believe correct.

src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
thomasgwatson: Tibet, as I noted in Slack, we could head this off at the pass by just passing the farm `bio` field into the group description on group creation and then never referencing the bio field again.

thomasgwatson (approved):
lgtm

src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
lorenjohnson: Hopefully that's what happens and this can be made simpler then. It'll work either way which is nice.

src/components/Widget/RichTextWidget/RichTextWidget.js
lorenjohnson: Just for confirmation of that. This video will not appear as part of the Welcome Widget, just About:  Screen Shot 2022-05-05 at 9 57 46 PM

tibetsprague (approved):

PR #1201, opened 2022-05-12
Add url sanitization to at a glance links

### LINKED ISSUE #1198: Farm "AT A GLANCE" external URLs don't work
When I click on any of the URIs in under the “AT A GLANCE” section at the top of the Farm Group Detail for a generated Farm the URL is getting treated as a path into Hylo not as an external URL.

thomasgwatson:
closes #1198

### DIFF
--- a/src/components/Widget/AtAGlanceWidget/AtAGlanceWidget.js
+++ b/src/components/Widget/AtAGlanceWidget/AtAGlanceWidget.js
@@ -4,6 +4,7 @@ import FancyLink from 'components/FancyLink'
 import { getSocialMedia, getWebsite, getAtAGlance } from 'store/selectors/farmExtensionSelectors'
 import { FarmDetailSection } from '../FarmDetailsWidget/FarmDetailsWidget'
 import './AtAGlanceWidget.scss'
+import { sanitizeURL } from 'util/url'
 
 export default function AtAGlanceWidget ({ group }) {
   /*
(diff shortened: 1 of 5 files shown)


src/components/Widget/AtAGlanceWidget/AtAGlanceWidget.js
```diff
@@ -4,22 +4,23 @@ import FancyLink from 'components/FancyLink'
...
+  const socialMedia = sanitizeURL(getSocialMedia(group))
+  const website = sanitizeURL(getWebsite(group))
   const atAGlance = getAtAGlance(group)
 
   return (
     <div styleName='at-a-glance-container'>
       {atAGlance.length > 0 && <FarmDetailSection items={atAGlance.map((item) => capitalize(item))} />}
-      {website && <FancyLink iconName='Website' linkUrl={website} title={`Website for ${group.name}`} />}
-      {socialMedia && <FancyLink iconName='Socials' linkUrl={socialMedia} title={`Social media link for ${group.name}`} />}
+      {website && <FancyLink iconName='Website' target='_blank' linkUrl={`https://${website}`} title={`Website for ${group.name}`} />}
```
tibetsprague, line 22: hmm, we always want to add https:// here? that seems unlikely that it will always be missing. won't sanitizeURL add it if we need it?

thomasgwatson: That was meant to be removed. But the strings that cause a problem are actually still valid URLs (eg: `computer.com` ). So really it is a weird quirk of our application that they are not handled as external links.

Ok, pushed a fix

tibetsprague: I think that must just be how react router works, it requires the protocol

PR #1227, opened 2022-06-01
Redirect /public/groups to login screen

tibetsprague:
Also do the work needed to have the GroupExplorer display at that URL while not logged in, but needs more thought to decide if we want that and more styling to make it look ok.

Fixes: https://github.com/Hylozoic/hylo-evo/issues/1225


thomasgwatson (approved):
lgtm

PR #1229, opened 2022-06-06
1207 Route to App when Logged In Fix

lorenjohnson:
The root cause of this issue was `userFetchedMiddleware` failing in the context of our SSR express setup, and causing the check for a currentUser to fail in that context.  That wasn't always the case, and not sure what of recent changes caused this, but the Mixpanel library isn't tolerant of not having a `document` object and hasn't been. It was failing silently before, but Mixpanel was still succeeding at identify and tracking.

This PR moves mixpanel out of redux and relies instead on the implicit window global scope of the `mixpanel` object from `mixpanel-browser`. Additionally, I eliminated the `userFetchedMiddleware` and moved the Mixpanel user identification steps into `AuthLayoutRouter`, making things a bit more clear and explicit. This should also help make it things a little more clear where to go once we get to GDPR stuff and wish to make tracking optional per-user. 

Currently deployed to Staging for QA.


tibetsprague (approved):
nice!

lorenjohnson:
Just did an edit in the description which made things a bit more clear. Also note that this PR adds a new env var `HYLO_COOKIE_NAME` and I have set the appropriate values on both staging and in production, in anticipation of this release.  Added this to the `.env.example` now as well.

thomasgwatson (approved):
lgtm

PR #1295, opened 2022-09-28
Evo 1284 switch topics to stream

### LINKED ISSUE #1284: Add search/filter to views
- [x] Add new Search tool to StreamViewControls. Clicking a magnifying glass button will open up a search bar like at the top of the Map drawer, except without all the topics options shown. You can add search terms by entering one and hitting enter and the search bar will disappear and the search term will show at the top of the stream and the stream will be filtered by that term. 
- [x] Get rid of the Feed, FeedList and all Feed components and use the Stream component everywhere (rename FeedBanner to StreamBanner?). Right now Topics use FeedList, as does the Projects view. And the Events view might also want to be merged into Stream? Maybe. 
- [x] Relatedly add StreamViewControls to Projects view and CustomViews, so people can change the View mode/style and sort for those views if they want. Default to the current settings. For Custom Views the Post Type filter selector will only show Post Types that are being shown on the Custom View.

thomasgwatson:
closes #1284


tibetsprague (commented):
LGTM

thomasgwatson:
@tibetsprague your comment didn't register as an approval :(

tibetsprague (approved):

PR #1297, opened 2022-10-05
Finish up oAuth login/consent flow

tibetsprague:
Allow for logged in users to still go through login/consent flow based on oAuth parameters. e.g. show the consent page always when prompt = consent

Fix/improve display when an app is asking for offline_access


src/routes/OAuth/Login/Login.js
```diff
@@ -0,0 +1,98 @@
...
+        window.location.href = redirectTo
+      } else {
+        dispatch(push('/'))
+      }
+    } catch (e) {
+      setError(e.message)
+    }
+  }
+
+  // TODO: let people log in with Google to login with Hylo? 🤪
```
thomasgwatson, line 43: if we do... might just be easier to do this as a routing redirect, no? Assuming that once they are logged in, they can just follow the logic denoted on line 24.

thomasgwatson (approved):
lgtm

src/routes/OAuth/Login/Login.js
tibetsprague: yes if they are already logged in they just go through the consent flow. now that i think about it we probably do need log in with google here in case they originally signed up to Hylo with Google. it just feels really weird to have someone go through the Google oAuth flow only to then go through ours... 🤔

tibetsprague: i think ill leave it this way unless/until it becomes an issue

Issue #1069, opened 2022-01-19
Remove SSR setup and deprecate our fork of babel-plugin-react-css-modules


lorenjohnson:
We use a our own for of a CSS Modules babel plugin which was created to support Server-Side Rendering (SSR) (i.e. this line [in `package.json`: `"babel-plugin-react-css-modules": "levity/babel-plugin-react-css-modules#ab83008",](https://github.com/Hylozoic/hylo-evo/blob/bd846b01d1a9118022262a2bfe881796c0027b63/package.json#L86)`)

This fork has been blocking or significantly complicating our upgrade path for Babel and related dependencies.  We should do what is required to get off this now very-out-of-date fork and move to the current version published in npm.

A key clue: https://github.com/gajus/babel-plugin-react-css-modules/issues/69. We initially forked for SSR-related reasons due to not being able to use WebPack in the backend render. 

I believe our SSR setup is defunct and adds unnecessary complexity in the initial setup layers of Evo, including the front-end express server setup. IMHO our SSR setup needs to be removed as part of this work, and will have the benefit of simplifying our setup.

⚠️ There will also likely be some implications to our overall PostCSS and Babel config, and some needed line-by-line global amendments to our SCSS files may be necessary (see also https://github.com/gajus/babel-plugin-react-css-modules/issues/49 for possible hints on that)

- [ ] Remove deprecated SSR setup
- [ ] Upgrade to latest React relevant Babel version including all relevant plugins (including `babel-plugin-react-css-modules`)
- [ ] Try it and churn until everything works
- [ ] Test
- [ ] Be happy, you're now a hero

lorenjohnson:
...probably replace all our current PostCSS + SASS config with: https://github.com/madyankin/postcss-modules

PR #1321, opened 2022-10-21
1: Clean-up and notes of Server-Side Rendering, Root `App` component, and Redux setup

lorenjohnson:
⚠️ Now includes these PRs: #1324 and #1330 

---

This is on the way to a few things we want: 

1. Reducing size and load time of `HyloMobileEditor` for HyloApp (a subsequent PR will be made off this branch to do that)
2. Ironing out some loops and confusing things in our Redux store setup. This will be useful when we move towards a shared store with Mobile, making decisions/amendments around ReduxORM, etc. Also, just generally makes the setup easier to grok and more standard. 
3. Making our very minimal and debatably beneficial server-side rendering (SSR) setup more clear and out of the way, giving us an easy way to address the forked Babel CSS Module WebPack stuff described in #1069. 
4. May make React Testing Library setup and use a little more clear and simple once we get back to refining that.


tibetsprague (approved):
nothing jumps out at me here, though seems like worth chatting about before merging for me to understand better?

PR #1322, opened 2022-10-21
2: `React.lazy` load `HyloEditorMobile` for code-splitting to improve load time in HyloApp

lorenjohnson:
**Remaining tasks:**

- [x] Remove store/dispatch dependency from mention and topic suggestions queries (this mobile editor bundle will not include the store). See notes to this regard in [the related HyloApp ticket](https://github.com/Hylozoic/HyloReactNative/issues/572)

**Benefits:**

* Reduces the size of `hyloApp/editor` / `HyloEditorMobile` load in HyloApp from ~2 MB to ~600k, which made a palpable difference in testing. It would be more important of a difference in the case of a poor or slow mobile data connection.

* Most likely we will also be creating a WebView for our ReactPlayer usage to get Video Emebeds working in the App. It will also want/need this top-level treatment: https://github.com/Hylozoic/HyloReactNative/issues/567

* A good and likely necessary first step towards moving the Mobile Editor code locally into the native app, which is where we ultimately should be (ref. https://github.com/Hylozoic/HyloReactNative/issues/572)

* Sets the stage for further lazy-loading / code-splitting down the line in the router. 

**Background:**

We are currently using `React.lazy` and `Suspense` to delay load of the `Notifications` and `Messages` drop-downs and this extends that usage to the very top. Our previously somewhat woven-in SSR setup wouldn't have allowed us to use `React.lazy` in non-authed views (`Suspense` is not supported by React server rendering). The cleaned-up SSR setup in the branch this is based upon should remedy that such that `App` (which was `clientRouter`) now will only ever get rendered client-side.


src/util/graphql.js
```diff
@@ -1,6 +1,9 @@
+import fetch from 'isomorphic-fetch'
```
thomasgwatson, line 1: Oh are we still specifically importing fetch? I think I used it 'raw' in a reactions PR... which might be a gap on older browsers (has fetch crossed the line into availability??)

thomasgwatson (approved):

tibetsprague:
so it this no longer a Draft?

lorenjohnson:
It likely needs to have `dev` merged-up and make sure things are still happy, but yeah it should be good.

PR #1334, opened 2022-10-27
Make sure bold text is bold across the site

tibetsprague:
All <strong> tags now use font Circular Bold since otherwise font-weight doesnt do anything. This is mainly to fix bold text in posts and comments but does apply everywhere.

Fixes https://github.com/Hylozoic/hylo-evo/issues/1333

### DIFF
--- a/src/css/typography.scss
+++ b/src/css/typography.scss
@@ -40,6 +40,10 @@ ul, ol {
   padding-left: 2em;
 }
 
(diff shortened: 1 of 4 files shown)


lorenjohnson (approved):
Yay. Much appreciate the extra diligence in checking other places. Paying down the debt...

Issue #1289, opened 2022-09-20
UX for switching groups needs a fix


cpolitano:
bug report / request > I continue to find it super disorienting when i switch between groups, and I land on the same view I was on before (Events, Projects, etc) instead of starting over at the Stream. it's especially a problem on the Custom Views - now if I'm looking at a custom view and I switch to a new group, the router points me to /groups/new-group/custom/1 which doesn't exist for most groups! this happened during a demo yesterday! can we please remove this UX?

Desired behavior: when I switch from one group to another, I land on the Stream for the new group

PR #1335, opened 2022-10-27
Always go to Stream (group home) when navigating to a new group or context

tibetsprague:


tibetsprague:
fixes https://github.com/Hylozoic/hylo-evo/issues/1289

lorenjohnson:
Curious more then concerned: Does this still maintain Map context when switching views?

src/util/navigation.js
```diff
@@ -51,21 +51,6 @@ export function baseUrl ({
   }
 }
 
-export function contextSwitchingUrl (newParams, routeParams) {
```
thomasgwatson, line 54: yaya deleting code!

thomasgwatson (approved):
lgtm

tibetsprague:
decided not to do the work to have the map stay on the map. i think lets try it this way and see if we get feedback from folks about it....

PR #1344, opened 2022-11-03
Fix canModerate in PostEditor

### LINKED ISSUE #1325: `canModerate` not accurate in PostEditor. Allows unauthorized creation of Announcements
Only the `currentGroup` value from Redux is used to determine `canModerate` vs actual selected group/s in `PostEditor`. `canModerate` in the editor is only currently used to determined whether you can post an announcement.

A couple of consequences of this:

* ⚠️ Opening the `Create` / PostEditor modal from a group you are a moderator of, then switching the "Post To" to any other group you are a member of will allow you to create a Post with "Announcement" in that group. 
* Conversely starting a post from outside of a group you are a moderator of (or from All or Public context) will never give you the Announcement option.

The fix here is to remove `canModerate` out of the connector logic and only derive it within the component from the actual group/s selected in the Post Editor.

thomasgwatson:
closes #1325

### DIFF
--- a/src/components/PostEditor/PostEditor.connector.js
+++ b/src/components/PostEditor/PostEditor.connector.js
@@ -64,7 +64,6 @@ export function mapStateToProps (state, props) {
   const topic = getTopicForCurrentRoute(state, props)
   const topicName = get('name', topic)
   const announcementSelected = state[MODULE_NAME].announcement
-  const canModerate = currentUser && currentUser.canModerate(currentGroup)
   const location = get('location', props)
   const postType = getQuerystringParam('newPostType', null, props)
   const isProject = postType === 'project' || get('type', post) === 'project'
@@ -72,7 +71,6 @@ export function mapStateToProps (state, props) {
 
   return {
     announcementSelected,
-    canModerate,
     context,
     currentGroup,
     currentUser,
--- a/src/components/PostEditor/PostEditor.js
+++ b/src/components/PostEditor/PostEditor.js
@@ -518,6 +518,17 @@ export default class PostEditor extends React.Component {
     })
   }
 
+  canModerate = () => {
+    const { myModeratedGroups = [] } = this.props
+    const { post } = this.state
+    const { groups = [] } = post
+    const myModeratedGroupsSlugs = myModeratedGroups.map(group => group.slug)
+    for (let index = 0; index < groups.length; index++) {
+      if (!myModeratedGroupsSlugs.includes(groups[index].slug)) return false
+    }
+    return true
+  }
+
   render () {
     const {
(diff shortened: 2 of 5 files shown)


src/components/PostEditor/PostEditor.connector.js
```diff
@@ -64,15 +64,13 @@ export function mapStateToProps (state, props) {
   const topic = getTopicForCurrentRoute(state, props)
   const topicName = get('name', topic)
   const announcementSelected = state[MODULE_NAME].announcement
-  const canModerate = currentUser && currentUser.canModerate(currentGroup)
```
tibetsprague, line 67: hmm, why is this not working? we use this function all over the app, why is it not working here? seems better to use common shared code than do something unique here

thomasgwatson: The use case it didn't work for here is that the user needs to be able to moderate all groups associated with a post, not just the original one that was selected when the PostEditor was opened.

Since the state of the PostEditor can change, the check for this was pushed down to that level, where it has access to the groups selected for the post. The post can only be an announcement if ALL groups selected are moderated by the user.

tibetsprague:
Ahhh yes, thanks

src/components/PostEditor/PostEditor.connector.js
tibetsprague: sooo, what should actually happen here? I actually think you should be able to choose announcement still, but the announcement will only get sent to groups the user moderates. @lorenjohnson what do you think? in which case this issue gets pushed back to the back-end. Though we still need to check whether any groups can be moderated to decide whether to show the announcement button, so this change is still needed

thomasgwatson: Yeah, let me know what the group consensus is here. @lorenjohnson if you have a view?

lorenjohnson: I would either only allow announcement in the case of a single group selection in which the user is a moderator, or allow it if the user is a moderator of all the selected groups. Either one I'm fine with.

Showing the announcement option for a mixed set of groups currently I think might be a little confusing or misleading, without a tiny bit more thought/work on the UI feedback around the feature. I don't think it's worth that at this moment and can be revisited if we get feedback or need from group moderators or otherwise we are really feeling it.

lorenjohnson: It looks like "allow if the user is a moderator of all the selected groups" is what you've done here @thomasgwatson, so I'm going to go ahead give give my own PR approval :)

lorenjohnson (approved):
This makes sense to me 👍🏼

PR #1345, opened 2022-11-03
Fixes #1343 - Fix formatting and autolink group descriptions

### LINKED ISSUE #1343: Group Description formatting is off and links are not "auto linked"
e.g. https://hylo.com/groups/collaborative-technology-alliance/explore/group/collaborative-technology-alliance

Need to turn "autolinking" for markdown content back on (in `hylo-shared#TextHelpers.markdown()`). 

** Also, should move the processing of this and `Message#text` bits into the backend model, and make sure to be sanitizing the output there.

lorenjohnson:
This will give proper formatting back to Group Descriptions as well as linking. 

* Also applies `ClickCatch` wherever markdown is used currently (so also on Message#text), as with `autolinking` on that is needed to keep folks from navigating away from Hylo when they click in links in those other contexts, as well as to keep Hylo links staying on the site. 

* Not required, but would be best to merge the backend Hylo-shared nudge soon so there isn't drift: https://github.com/Hylozoic/hylo-node/pull/886

### DIFF
--- a/src/components/GroupCard/GroupCard.js
+++ b/src/components/GroupCard/GroupCard.js
@@ -1,13 +1,14 @@
 import React from 'react'
-import cx from 'classnames'
+import { Link } from 'react-router-dom'
+import { capitalize } from 'lodash'
 import { TextHelpers } from 'hylo-shared'
-import './GroupCard.scss'
+import { groupUrl, groupDetailUrl } from 'util/navigation'
+import ClickCatcher from 'components/ClickCatcher'
 import GroupHeader from './GroupHeader'
 import HyloHTML from 'components/HyloHTML'
-import { Link } from 'react-router-dom'
-import { groupUrl, groupDetailUrl } from 'util/navigation'
 import Pill from 'components/Pill'
-import { capitalize } from 'lodash'
+import cx from 'classnames'
+import './GroupCard.scss'
 
 /*
   Each card needs
@@ -52,7 +53,9 @@ export default function GroupCard ({
         />
         {group.description
           ? <div styleName='group-description'>
-            <HyloHTML element='span' html={TextHelpers.markdown(group.description)} />
+            <ClickCatcher groupSlug='all'>
+              <HyloHTML element='span' html={TextHelpers.markdown(group.description)} />
+            </ClickCatcher>
           </div>
           : ''
         }
--- a/src/components/GroupCard/GroupCard.js
+++ b/src/components/GroupCard/GroupCard.js
@@ -53,7 +53,7 @@ export default function GroupCard ({
         />
         {group.description
           ? <div styleName='group-description'>
-            <ClickCatcher groupSlug='all'>
+            <ClickCatcher>
               <HyloHTML element='span' html={TextHelpers.markdown(group.description)} />
             </ClickCatcher>
           </div>
--- a/src/components/Widget/GroupsWidget/GroupsWidget.js
+++ b/src/components/Widget/GroupsWidget/GroupsWidget.js
@@ -3,9 +3,10 @@ import PropTypes from 'prop-types'
 import { Link } from 'react-router-dom'
 import Slider from 'react-slick'
 import { TextHelpers } from 'hylo-shared'
-import HyloHTML from 'components/HyloHTML'
-import { DEFAULT_BANNER, DEFAULT_AVATAR } from 'store/models/Group'
 import { createGroupUrl, groupUrl, groupDetailUrl } from 'util/navigation'
+import { DEFAULT_BANNER, DEFAULT_AVATAR } from 'store/models/Group'
+import ClickCatcher from 'components/ClickCatcher'
+import HyloHTML from 'components/HyloHTML'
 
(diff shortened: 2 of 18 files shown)


src/components/GroupCard/GroupCard.js
```diff
@@ -52,7 +53,9 @@ export default function GroupCard ({
         />
         {group.description
           ? <div styleName='group-description'>
-            <HyloHTML element='span' html={TextHelpers.markdown(group.description)} />
+            <ClickCatcher groupSlug='all'>
```
tibetsprague, line 56: why is groupSlug always all, shouldn't it be the group slug?

src/components/Widget/GroupsWidget/GroupsWidget.js
```diff
@@ -67,7 +68,9 @@ export function GroupCard ({ group, routeParams, className }) {
           <div styleName='group-name'>{group.name}</div>
           <div styleName='member-count'>{group.memberCount} member{group.memberCount !== 1 ? 's' : ''}</div>
           <div styleName='group-description'>
-            <HyloHTML element='span' html={TextHelpers.markdown(group.description)} />
+            <ClickCatcher groupSlug='all'>
```
tibetsprague, line 71: same here

tibetsprague (commented):
just wondering why groupSlug is all and not the actual group slug

src/components/GroupCard/GroupCard.js
lorenjohnson: Sorry it's a long description, but all true: 

My goal in pushing this change is simply to get external links handled in the group description the same way they are handled everywhere else. 

It will only come-up if someone puts in a mention or a topic in the description, which there basically isn't an obvious way for a user to do right now. 

The Group Description entry probably now be given a `HyloEditor` instead of a plain text markdown interpreted input, but I didn't want to get into that now either and it wasn't something I was trying to fix with this. 

\* Putting the base slug as that of the group being described means that in many cases the user won't yet be a member of that group at all, so if there were a mention or a topic linked here in the description tied to that group context, it could be inaccessible to the user viewing the description when it maybe wouldn't be otherwise? 

Assuming `currentGroup` would probably be arbitrary to this content, so not that. 

The default for the `ClickCatcher` in its few pieces is 'all' anyway for topics and mentions and I think that is all the slug is used for in the catcher. Outsides links don't care, and relative Hylo links will just be the actual path with no regard for this prop. So actually the `groupSlug` prop should just be left off in all these cases and added if/when we find cases where we want a different result down the road. For now I included as it felt safer and more explicit to do so.

lorenjohnson: Furthermore, and I need to create a ticket for this, the markdown transformations happening client side here and with Messages#text should all be moved to the backend and sanitized on the way out. I imagined that work happening at the same time we get around to dropping `HyloEditor` in as the input. It will need to be thought through then what we want to do with the existing markdown content as the new or edited entries will be stored as HTML once we are using the editor. Probably we'll just materialize the previous content to HTML in a migration.

lorenjohnson: FYI - created these two follow-up tickets which maybe should happen in one fell swoop but would be fine to one-by-one as well: 

https://github.com/Hylozoic/hylo-evo/issues/1349
https://github.com/Hylozoic/hylo-evo/issues/1350

src/routes/GroupDetail/GroupDetail.js
```diff
@@ -196,8 +200,9 @@ export class UnwrappedGroupDetail extends Component {
             </div>
           ) : (
             <div styleName='g.groupDescription'>
-              <GroupAboutVideoEmbed uri={group.aboutVideoUri} styleName='g.groupAboutVideo' />
```
lorenjohnson, line 199: * Despite that this is moved out of the `div.groupDescription` it won't have a styling impact as there is no top margin or padding on `groupDescription`. There is a bottom margin on `aboutVideoUri` which will remain where it was before above the description, and now above the "enter a description" message (which is good).

src/components/GroupCard/GroupCard.js
tibetsprague: gotcha, makes sense!

tibetsprague (approved):

lorenjohnson:
@tibetsprague I went ahead and removed the explicit `groupSlug='all'` on all instances of `ClickCatcher` where that was specified, and added explicitly added it as the default for the prop on `ClickCatcher`.

I confirmed that in all cases the default `groupSlug` down stream (currently only `topicPath` and `mentionPath`, ref. https://github.com/Hylozoic/hylo-shared/blob/master/src/PathHelpers.js) was already `all`, but added it as the default component prop as well as I think that is also the business of `ClickCatcher`. Ultimately I think this communicates best. 

Hope that makes sense, and either way this still won't change what happens currently in any case. 

The issue with contextualizing `ClickCatcher` is not something we're addressing here, but will be given the right attention in those cited follow-up tickets.

Also, added the notes about this all to the `README`. Going to go ahead and merge here.

Issue #1352, opened 2022-11-06
Fix `history` / `history.goBack` behavior for `LandingPage` "About us" modal


lorenjohnson:
This is a necessary change for the App so I can properly handle the go back functionality in the WebView when a user is on "Explore", goes to "About Us", and then goes back. The current behavior is using `navigation#removeGroupFromUrl` which after analysis and testing, I don't think is necessary. 

_Footnote: I think path munging at this level is always a code smell / at best a "worst case" which should send us back to reconciling our routing._

PR #1353, opened 2022-11-06
Fixes # 1352 back button behavior for LandingPage > "About Us"

lorenjohnson:
#1352

### DIFF
--- a/src/routes/GroupDetail/GroupDetail.js
+++ b/src/routes/GroupDetail/GroupDetail.js
@@ -1,7 +1,7 @@
 import cx from 'classnames'
 import { get, keyBy, map, trim } from 'lodash'
 import React, { Component, useState } from 'react'
-import { Link, useHistory } from 'react-router-dom'
+import { Link, useHistory, useLocation } from 'react-router-dom'
 import PropTypes from 'prop-types'
 import { TextHelpers, WebViewMessageTypes } from 'hylo-shared'
 import isWebView, { sendMessageToWebView } from 'util/webView'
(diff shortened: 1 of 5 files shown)


src/routes/GroupDetail/GroupDetail.js
```diff
@@ -86,25 +87,24 @@ export class UnwrappedGroupDetail extends Component {
...
 
-    const fullPage = !onClose
-
     return (
       <div className={cx({ [g.group]: true, [g.fullPage]: fullPage, [g.isAboutCurrentGroup]: isAboutCurrentGroup })}>
         <div styleName='g.groupDetailHeader' style={{ backgroundImage: `url(${group.bannerUrl || DEFAULT_BANNER})` }}>
-          {onClose && (
-            <a styleName='g.close' onClick={onClose}><Icon name='Ex' /></a>
+          {!fullPage && (
+            <a styleName='g.close' onClick={history.goBack}><Icon name='Ex' /></a>
```
tibetsprague, line 107: hmm, but what if there's no page to go back to? what if they loaded this page directly? or the page to go back to

lorenjohnson: You're entirely correct. I think I was trying to ignore that case. I have updated now to cover that case and tested it for all the places this routes matches: https://github.com/Hylozoic/hylo-evo/pull/1353/files#diff-a9d11da53b47e893e82fb72c59e6020964f5d124e484f9dbd4213a3183cc30c8R396-R402

Should be good to go.

lorenjohnson: I think the root issue here was potentially just that the `history.push` that was happening was via the `connect-react-router` API which may not have been populating the browser history object appropriately. That's my suspicion, and either way I consider that way of navigating legacy and we should use the React Router provided `history` directly. Whatever the case what is here is now doing what I need in the App.

lorenjohnson: Bonus note: I created this initially and it worked fine, BUT MapExplorer initial load includes a redirect from the Stream somehow so in that case going back would from a first-page-load of the "About us" would send the user back to the Stream instead of the Group Detail. Hurumph. I'd like to stay as close to browser native stuff as possible (in avoidance of duplicating state management up the stack)... we were so close :) 

```
import { useHistory } from 'react-router-dom'

export default function useSafeGoBack (defaultPath = '/') {
  const history = useHistory()

  return () => {
    if (document.referrer.startsWith(document.location.origin)) {
      history.goBack()
    } else {
      history.push(defaultPath)
    }
  }
}
``

tibetsprague (approved):

Issue #883, opened 2021-04-22
When accepting a join request the person should be added right away to the directory (or the directory should reload)


tibetsprague:

tibetsprague:
To test this you will need to setup a group that is Public and Protected, see settings here: https://localhost:3000/groups/terran-collective/settings

Then you will need to go to that group's page as a user who is not a part of that group and request to join. 
Then as a moderator of the group you go to https://localhost:3000/groups/terran-collective/settings/requests and Accept the join request. 
Then navigate to the Member Directory and you'll see that the new person who joined is not listed in the member directory. 

What needs to happen is that is that in when the ACCEPT_JOIN_REQUEST (see src/routes/GroupSettings/MembershipRequestsTab/MembershipRequestsTab.store.js) redux action is completed the new person needs to be added to the members of the group in redux. 
This adding would happen in src/store/reducers/ormReducer/index.js by watching for ACCEPT_JOIN_REQUEST and from that join request looking up the group and the member ids and adding a new Membership to redux ORM with the member id and group id.

Sofacosta:
I tested it twice and it works for me 😬

tibetsprague:
Hmm I just tested myself and it didn’t work. This bug is that they should appear right away in the directory without having to refresh the page. On refresh yes the new person appears. For you they are appearing immediately?

PR #1360, opened 2022-11-10
Scroll to comment, based on url

thomasgwatson:
Closes the EVO side of Node ticket [#883](https://github.com/Hylozoic/hylo-node/issues/883)

### DIFF
--- a/src/routes/PostDetail/Comments/Comment/Comment.js
+++ b/src/routes/PostDetail/Comments/Comment/Comment.js
@@ -5,7 +5,7 @@ import ReactDOM from 'react-dom'
 import { Link } from 'react-router-dom'
 import { filter, isFunction } from 'lodash/fp'
 import { TextHelpers } from 'hylo-shared'
-import { personUrl, findCommentId } from 'util/navigation'
+import { personUrl } from 'util/navigation'
 import scrollIntoView from 'scroll-into-view-if-needed'
 import ShowMore from '../ShowMore'
 import Tooltip from 'components/Tooltip'
@@ -85,7 +85,7 @@ export class Comment extends Component {
   }
 
   render () {
-    const { canModerate, comment, currentUser, deleteComment, onReplyComment, removeComment, slug } = this.props
+    const { canModerate, comment, currentUser, deleteComment, onReplyComment, removeComment, slug, selectedCommentId } = this.props
     const { id, creator, createdAt, text, attachments } = comment
     const { editing, scrolledToComment } = this.state
     const isCreator = currentUser && (comment.creator.id === currentUser.id)
@@ -96,7 +96,6 @@ export class Comment extends Component {
       { icon: 'Trash', label: 'Delete', onClick: isCreator ? () => deleteComment(comment.id) : null },
       { icon: 'Trash', label: 'Remove', onClick: !isCreator && canModerate ? () => removeComment(comment.id) : null }
     ])
-    const selectedCommentId = findCommentId(window.location.pathname)
     if (this.commentRef.current && selectedCommentId === comment.id && !scrolledToComment) this.handleScrollToComment()
 
     return (
--- a/src/routes/PostDetail/Comments/Comment/Comment.js
+++ b/src/routes/PostDetail/Comments/Comment/Comment.js
@@ -1,10 +1,12 @@
+import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React, { Component } from 'react'
 import ReactDOM from 'react-dom'
 import { Link } from 'react-router-dom'
 import { filter, isFunction } from 'lodash/fp'
 import { TextHelpers } from 'hylo-shared'
-import { personUrl } from 'util/navigation'
+import { personUrl, findCommentId } from 'util/navigation'
+import scrollIntoView from 'scroll-into-view-if-needed'
 import ShowMore from '../ShowMore'
 import Tooltip from 'components/Tooltip'
 import Avatar from 'components/Avatar'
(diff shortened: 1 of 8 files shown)


src/routes/PostDetail/Comments/Comment/Comment.js
```diff
@@ -76,9 +96,11 @@ export class Comment extends Component {
       { icon: 'Trash', label: 'Delete', onClick: isCreator ? () => deleteComment(comment.id) : null },
       { icon: 'Trash', label: 'Remove', onClick: !isCreator && canModerate ? () => removeComment(comment.id) : null }
     ])
+    const selectedCommentId = findCommentId(window.location.pathname)
```
tibetsprague, line 99: shouldn't this be availing in routeParams? seems slightly off to search through window.location.pathname, especially here in the render (also could be passed in from the connector)

src/routes/PostDetail/Comments/Comments.connector.js
```diff
@@ -2,16 +2,22 @@ import { connect } from 'react-redux'
 import getMe from 'store/selectors/getMe'
 import { isEmpty } from 'lodash/fp'
 import fetchComments from 'store/actions/fetchComments'
+import { findCommentId } from 'util/navigation'
```
tibetsprague, line 5: not used? unless we want to load it here and pass down, though again should be in routeParams I think and you could use getRouteParam

src/routes/PostDetail/Comments/Comments.js
```diff
@@ -29,6 +28,26 @@ export default class Comments extends Component {
     comments: []
   }
 
+  componentDidMount () {
+    this.ensureSelectedCommentPresent()
+  }
+
+  componentDidUpdate (prevProps) {
+    this.ensureSelectedCommentPresent()
```
tibetsprague, line 36: why do we want to do this every time? seems like we should at least put this behind a check if certain prop(s) changed

src/routes/PostDetail/Comments/Comments.js
```diff
@@ -29,6 +28,26 @@ export default class Comments extends Component {
...
+
+  ensureSelectedCommentPresent () {
+    const { comments, commentsPending } = this.props
+    const selectedCommentId = findCommentId(window.location.pathname)
+    const commentIds = []
+    comments.forEach(comment => {
+      commentIds.push(comment.id)
+      comment.childComments.forEach(comment => commentIds.push(comment.id))
+    })
+    if (comments.length < 1) return
```
tibetsprague, line 47: should probably go at the top of the function

src/routes/PostDetail/Comments/Comments.js
```diff
@@ -29,6 +28,26 @@ export default class Comments extends Component {
...
+    if (!commentsPending && !commentIds.includes(selectedCommentId.toString())) this.props.fetchComments().then(() => this.forceUpdate())
```
tibetsprague, line 48: so weird that it wouldn't re-render. maybe something is not being passed down from the connector that should be? the comments should change and then a render should happen

src/util/navigation.js
```diff
@@ -109,7 +109,7 @@ export function editPostUrl (id, opts = {}, querystringParams = {}) {
 }
 
 export function commentUrl (postId, commentId, opts = {}, querystringParams = {}) {
-  return `${postUrl(postId, opts, querystringParams)}#comment_${commentId}`
```
tibetsprague, line 112: oh interesting, this already existed. was it doing anything anywhere? I know people do use this # strategy to go to specific places on a page sometime, I actually now wonder if we should stick with that? but dont have a strong feeling. might look into why people use # like this sometimes. and make sure this isn't being used anywhere else in the code

src/util/navigation.js
```diff
@@ -185,3 +185,9 @@ export const origin = () =>
 export function isPublicPath (path) {
   return (path.startsWith('/public'))
 }
+
+export function findCommentId (url) {
```
tibetsprague, line 189: yeah seems like we should usually (always?) use the routeParam instead of a manual thing like this. unless we are using the #

src/routes/PostDetail/Comments/Comment/Comment.js
thomasgwatson: I checked `routeParams` and it wasn't there. I am assuming that it needs to be specifically declared in our routes for React Router to pick it up as a param, which involves some more optional matching and complicates routing even further (which perhaps I'm just about to get stuck into with Loren anyway). But ultimately, why feed it through routeParams, or the connector, when you can just get it straight from the source, without using additional libraries etc?

src/routes/PostDetail/Comments/Comments.js
thomasgwatson: the checks are contained in the called function

src/routes/PostDetail/Comments/Comments.js
thomasgwatson: Indeed.

src/util/navigation.js
thomasgwatson: As far as I could find, it wasn't doing anything. From what I briefly read, using the hash is mimicing the pattern of denoting various headers with a hash identifier, so that the browser could navigate to different sections of a page. In our case, we have specific entities with known IDs, so its probably considered more in line with RESTful urls to have it part of the path

src/routes/PostDetail/Comments/Comment/Comment.js
tibetsprague: yes you would need to add the optional matching. that seems more idiomatic and aligned with how we currently do things and how routing is supposed to work to me... @lorenjohnson curious your thoughts here

tibetsprague (approved):

PR #1372, opened 2022-11-16
1301 chat rooms

tibetsprague:


src/components/EmojiPicker/EmojiPicker.js
```diff
@@ -23,6 +23,7 @@ export default function EmojiPicker (props) {
       handleReaction(selectedEmoji)
     }
     setModalOpen(!modalOpen)
+    return true
```
thomasgwatson, line 26: whats this do?

src/routes/ChatRoom/ChatPost/ChatPost.js
```diff
@@ -0,0 +1,254 @@
...
+    imageAttachments,
+    linkPreview,
+    linkPreviewFeatured,
+    myReactions,
+    postReactions
+  } = post
+
+  if (intersectionObserver) {
+    useEffect(() => {
+      intersectionObserver.observe(ref.current)
```
thomasgwatson, line 70: Whats this do?

src/routes/ChatRoom/ChatPost/ChatPost.js
```diff
@@ -0,0 +1,254 @@
...
+  const removePostWithConfirm = useCallback((event) => {
+    if (window.confirm('Are you sure you want to remove this post? You cannot undo this.')) {
+      dispatch(removePost(id, group.slug))
+    }
+    event.stopPropagation()
+    return true
+  })
+
+  const actionItems = filter(item => isFunction(item.onClick), [
+    // { icon: 'Pin', label: pinned ? 'Unpin' : 'Pin', onClick: pinPost },
```
thomasgwatson, line 147: these ones still need to be commented out?

src/routes/ChatRoom/ChatPost/ChatPost.js
```diff
@@ -0,0 +1,254 @@
...
+        {details && (
+          <ClickCatcher groupSlug={group.slug}>
+            {editing
+              ? <HyloEditor
+                contentHTML={details}
+                groupIds={post.groups.map(g => g.id)}
+                // onAddTopic={handleAddTopic}
+                onEscape={handleEditCancel}
+                onEnter={handleEditSave}
+                // onUpdate={handleDetailsUpdate}
```
thomasgwatson, line 207: Would have thought chats still need this tho?

src/routes/ChatRoom/ChatRoom.connector.js
```diff
@@ -0,0 +1,224 @@
...
+      .filter(x => includes(x.id, results.ids))
+      .orderBy(p => Number(p.id))
+      .toModelArray()
+      .map(p => presentPost(p))
+  }
+)
+
+const NUM_POSTS_TO_LOAD = 25
+
+export function mapStateToProps (state, props) {
```
thomasgwatson, line 60: So you have a connector for Chatroom, using the class component style of redux, feeding into Chatroom.js, which is a functional component? Mixing these patterns has only ever given me huge grief in the past. I really recommend avoiding the use of the connector pattern directly feeding into a functional component.

src/routes/ChatRoom/ChatRoom.js
```diff
@@ -0,0 +1,514 @@
...
+  const ref = useRef()
+  useLayoutEffect(() => {
+    ref.current = fn
+  })
+  return useCallback(() => (0, ref.current)(), [])
+}
+
+export default function ChatRoom (props) {
+  const {
+    // changeSearch,
```
thomasgwatson, line 47: commented out for now? I don't know if search will be a feature for chat

thomasgwatson (approved):
lgtm, a lot going on, thats for sure. If you are still having issues with the connector feeding into the functional component for the Chatroom, I suspect you either need to memoize everything in the connector with something like reselect or convert the connector content into hooks

src/components/PostCard/PostDetails/PostDetails.scss
```diff
@@ -54,6 +54,23 @@
   font-size: 16px;
   line-height: 24px;
   margin-bottom: 20px;
+
```
lorenjohnson, line 57: Wondering what happens in the case of `PostCard` for these same styles? Can these styles be in `global/typography.scss`, or otherwise be reconciled there if these are now importantly specific to Post Details?

src/routes/ChatRoom/ChatPost/ChatPost.js
```diff
@@ -0,0 +1,254 @@
...
+            </div>
+          </>
+        )}
+        {details && (
+          <ClickCatcher groupSlug={group.slug}>
+            {editing
+              ? <HyloEditor
+                contentHTML={details}
+                groupIds={post.groups.map(g => g.id)}
+                // onAddTopic={handleAddTopic}
```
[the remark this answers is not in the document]
lorenjohnson: I can answer that one :) The `onAddTopic` callback is used in the PostEditor to populate the topics field which chats don't utilize. The line can just be deleted vs commented-out though.

src/routes/ChatRoom/ChatPost/ChatPost.js
lorenjohnson: Same here I think. This is a callback which in the PostEditor context is going to be used to scan entry as it's happening, and in the Chat entry context it's evidently not needed. Delete line.

src/routes/ChatRoom/ChatRoom.connector.js
lorenjohnson: In agreement on this. I think we should consider `.connector.js` files deprecated in the case of new components. I haven't personally had any issue with having a connector HOC wrapping a functional component which also uses redux hooks, but I do think it's hard to track/understand what's going on.

src/routes/ChatRoom/index.js
```diff
@@ -0,0 +1,4 @@
+import ChatRoom from './ChatRoom'
```
lorenjohnson, line 1: Small thing: We most consistently make this generic in our code, i.e. `import component from './ChatRoom'`

src/store/reducers/ormReducer/index.js
```diff
@@ -588,3 +639,19 @@ export default function ormReducer (state = orm.getEmptyState(), action) {
 
   return session.state
 }
+
+// XXX: this is ugly, would be better to load these posts through redux-orm "queries" so they update automatically
```
lorenjohnson, line 643: Not action for now--just a note: 

This is a tangle of things, but zoomed-out a bit I think it would be better mutate and rely upon the return from that mutation to make the update vs doing things optimistically. Adding/preferring a brief loading indicator until the result of the mutation comes back and updates through the Model Extractor... The amount of futzing required to do optimistic updates right now I think is creating debt which will make a transition away from our particular use of ReduxORM and such more work.

lorenjohnson (changes requested):
Some notes. Nothing critical. Could probably go over things again, but seems like there has been a decent amount of testing all around so... Fix things from notes if they resonate, and otherwise LGTM!

src/routes/ChatRoom/ChatRoom.connector.js
tibetsprague: yeah I think you are right, though actually I think ChatRoom should not have been a functional component, its too complex and the hooks are super confusing. not sure what to do at this point

src/routes/ChatRoom/ChatRoom.js
tibetsprague: yeah I could delete, not sure if we will add search to chat rooms either

src/components/PostCard/PostDetails/PostDetails.scss
tibetsprague: good questions, must be @brodeur it doing but I can also look at

src/store/reducers/ormReducer/index.js
tibetsprague: the only issue is that not doing optimistically makes the app feel slow compared to other chat apps. right now we dont add new chats optimistically but i think people will complain about it

PR #1373, opened 2022-11-17
Evo 1050 group page public

### LINKED ISSUE #1050: Public landing page for public groups (based on GroupDetail)
Should be able to go to the URL for a public group and view it without being signed in to Hylo. If the group is open or restricted you can request to join but then you have to sign in to Hylo first.

This displays the GroupDetail component, just like if a signed in user goes to a group URL they are not a member of yet

thomasgwatson:
closes #1050 

Note, posts have been switched off for now. If we want to also show public posts in the group page to non-logged in peeps (which sounds great), lets do it in another ticket

### DIFF
--- a/src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
+++ b/src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
@@ -12,7 +12,7 @@ export default function FarmGroupDetailBody ({
   routeParams
 }) {
   const { group } = useEnsureCurrentGroup()
-  const { posts } = (currentUser && useEnsurePosts({ public: true, sortBy: 'updated', context: 'groups' })) || { posts: [] }
+  const { posts } = useEnsurePosts({ public: true, sortBy: 'updated', context: 'groups', currentUser })
   const bio = getBio(group)
 
   // TODO: hide widgets if they have no data, means loading all the data here?
--- a/src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
+++ b/src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
@@ -12,7 +12,7 @@ export default function FarmGroupDetailBody ({
   routeParams
 }) {
   const { group } = useEnsureCurrentGroup()
-  const { posts } = useEnsurePosts({ public: true, sortBy: 'updated', context: 'groups' })
+  const { posts } = (currentUser && useEnsurePosts({ public: true, sortBy: 'updated', context: 'groups' })) || { posts: [] }
   const bio = getBio(group)
 
   // TODO: hide widgets if they have no data, means loading all the data here?
@@ -29,9 +29,9 @@ export default function FarmGroupDetailBody ({
     },
     { settings: {}, isVisible: true, name: 'farm_details' },
     { settings: {}, isVisible: true, name: 'opportunities_to_collaborate' },
-    { settings: {}, isVisible: true, name: 'relevant_requests_offers' },
-    { settings: {}, isVisible: true, name: 'relevant_project_activity' },
-    { settings: {}, isVisible: true, name: 'relevant_events' },
+    { settings: {}, isVisible: !!currentUser, name: 'relevant_requests_offers' },
+    { settings: {}, isVisible: !!currentUser, name: 'relevant_project_activity' },
+    { settings: {}, isVisible: !!currentUser, name: 'relevant_events' },
     { settings: {}, isVisible: true, name: 'farm_map' },
     { settings: {}, isVisible: true, name: 'farm_open_to_public' }
   ]
--- a/src/routes/GroupDetail/GroupDetail.connector.js
+++ b/src/routes/GroupDetail/GroupDetail.connector.js
@@ -49,7 +49,7 @@ export function mapDispatchToProps (dispatch, props) {
   return {
     addSkill: (name) => dispatch(addSkill(name)),
     removeSkill: (skillId) => dispatch(removeSkill(skillId)),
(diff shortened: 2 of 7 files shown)


src/routes/GroupDetail/GroupDetail.connector.js
```diff
@@ -49,7 +49,7 @@ export function mapDispatchToProps (dispatch, props) {
   return {
     addSkill: (name) => dispatch(addSkill(name)),
     removeSkill: (skillId) => dispatch(removeSkill(skillId)),
-    fetchGroup: () => dispatch(fetchGroupDetails({ slug, withWidgets: true })),
+    fetchGroup: (currentUser) => dispatch(fetchGroupDetails({ slug, withWidgets: true, withPrerequisites: !!currentUser })),
```
tibetsprague, line 52: you could also handle this on the backend entirely, but this is ok

src/routes/PublicLayoutRouter/PublicLayoutRouter.js
```diff
@@ -1,77 +1,195 @@
...
+    (async () => {
+      setLoading(true)
 
-  const pathMatchParams = useMemo(() => (
-    matchPath(location.pathname, [
-      '/:context(public)/:view(groups|map)?'
-    ])?.params || { context: 'public' }
-  ), [location.pathname])
+      const result = await dispatch(checkIsPublicGroup(groupSlug))
+      const isPublicGroup = result?.payload?.data?.group?.visibility === 2
```
tibetsprague, line 67: won't this error out if you cant see the group, or just return null? seems like you dont need this check.

src/routes/PublicLayoutRouter/PublicLayoutRouter.scss
```diff
@@ -18,7 +18,6 @@
 
 .background {
   width: 100%;
-  height: 100%;
```
tibetsprague, line 21: this seems like something to test, what was happening?

src/routes/RootRouter/RootRouter.js
```diff
@@ -41,51 +41,23 @@ export default function RootRouter () {
...
-        <Route path='/public/groups' exact component={NonAuthLayoutRouter} />
-        <Route path='/public' component={PublicLayoutRouter} />
-        <Route path={'(.*)' + POST_DETAIL_MATCH} component={CheckPublicPost} />
+        <Route
+          path='/:context(groups)/:groupSlug/join/:accessCode'
+          component={NonAuthLayoutRouter}
+        />
+        <Route
+          path={[
+            '/:context(public)/:view(map|groups)?', // so I will need to add something here to great public, non-auth group page up? and without stopping the group join mechanism detailed by Loren
```
lorenjohnson, line 53: I don't understand-- what are you saying that I'm saying that I think we need to do 😂 ?  In general to remember/confirm what should happen here I'd test the couple routes for signing-up (invite link) or token-auth'ing (email invite) to a join a group when not-authorized. If those flows work, we're good, if not we may need to adjust something here. On glance without digging into it now, I suspect this is all fine as is, but test it out if you're not sure.

src/components/FarmGroupDetailBody/FarmGroupDetailBody.js
```diff
@@ -12,7 +12,7 @@ export default function FarmGroupDetailBody ({
   routeParams
 }) {
   const { group } = useEnsureCurrentGroup()
-  const { posts } = useEnsurePosts({ public: true, sortBy: 'updated', context: 'groups' })
+  const { posts } = (currentUser && useEnsurePosts({ public: true, sortBy: 'updated', context: 'groups' })) || { posts: [] }
```
lorenjohnson, line 15: I would think that React would be unhappy to only sometimes run this hook, but I may be wrong. To be safe I'd probably assign this response out to something like `postsFromStore` or whatever, then ternary `postsFromStore?.posts` to return either that result or `[]`...

lorenjohnson: Alternatively, you could change `useEnsurePosts` to return a default of an empty array for the `posts` key...

src/routes/PublicLayoutRouter/PublicLayoutRouter.js
lorenjohnson: Sorry, strike that. I see what you mean @tibetsprague. If the query returns anything then its implicitly Public I guess. So yeah, you shouldn't also need to check for the visibility status at this point.

lorenjohnson (changes requested):
I am mostly approving this, but see notes. Also needs confirmation via finger-testing / QA to confirm that non-auth'd Group Joining works as expected. Also, I didn't run this to confirm layout and linking within the Public context. That is something I'd like to have confirmed is all good before proceeding to merge.

src/routes/PublicLayoutRouter/PublicLayoutRouter.js
thomasgwatson: No, it doesn't need it at all, its just being emphatic

src/routes/RootRouter/RootRouter.js
thomasgwatson: redundant comment, sorry

src/routes/GroupDetail/GroupDetail.connector.js
thomasgwatson: How so? The frontend was blowing up because it was making queries that expected a valid `currentUser` but that isn't assured in this context, so I just needed to adjust this part of the query to be included or not depending on the presence of a currentUser

And, didn't want to make a backend change if I didn't need to.

tibetsprague (approved):

lorenjohnson (approved):

PR #1380, opened 2022-11-23
1375 newly created posts not appearing in stream

tibetsprague:


lorenjohnson (approved):
I will need to sync this stuff to the App I think. Making a ticket to analyze whether the same issues are there.

PR #1381, opened 2022-11-23
1376 show post details from topic streams

tibetsprague:


src/routes/AuthLayoutRouter/AuthLayoutRouter.js
```diff
@@ -353,6 +353,7 @@ export default function AuthLayoutRouter (props) {
             {/* NOTE: These routes could potentially be simply: `(.*)/${POST_DETAIL_MATCH}` and`(.*)/${GROUP_DETAIL_MATCH}` */}
             <Switch>
               <Route path={`/:context(all|public)/:view(events|explore|groups|map|projects|stream)/${POST_DETAIL_MATCH}`} component={PostDetail} />
+              <Route path={`/:context(all|public)/:view(topics)/:topicName/${POST_DETAIL_MATCH}`} component={PostDetail} />
```
lorenjohnson, line 356: You probably already get this, but you could also done this way which has a couple of minor and subtle advantages in my book which we can weigh another time. Just highlighting the option/consideration for now:

```
<Route path={[
  `/:context(all|public)/:view(events|explore|groups|map|projects|stream)/${POST_DETAIL_MATCH}`,
   `/:context(all|public)/:view(topics)/:topicName/${POST_DETAIL_MATCH}`
  ]}
  component={PostDetail}
/>
```

Also, the note on line 353 remains mostly true. What is in this whole details section when broken down a little reduces to always rendering `PostDetail` when the end of the path matches `/post/:postId(HYLO_ID)`, and vice versa for `/groups/:groupSlug`. I think that's something worth doing at some point as it will make things a little more maintainable and make more clear what I think is one really important pattern to recognize in our routing... If that makes sense.

lorenjohnson (approved):
LGTM! Note is mostly academic at the moment, let's keep this moving as is!

src/routes/AuthLayoutRouter/AuthLayoutRouter.js
tibetsprague: yes totally, i would love to make that change, just will require more testing. maybe in the more significant routing stuff Tom and you have been working on we do that too

PR #1384, opened 2022-12-05
Node 884 clean up paths

thomasgwatson:
For https://github.com/Hylozoic/hylo-node/issues/884


src/routes/AuthLayoutRouter/AuthLayoutRouter.js
```diff
@@ -81,12 +81,16 @@ export default function AuthLayoutRouter (props) {
     `/(.*)/${POST_DETAIL_MATCH}`,
     `/(.*)/${GROUP_DETAIL_MATCH}`
   ])
+
+  const paramPostId = matchPath(location.pathname, [
```
tibetsprague, line 85: I wonder if in this case since we are using this again we can use @lorenjohnson suggestion of just .*/${POST_DETAIL_MATCH}?

src/routes/AuthLayoutRouter/AuthLayoutRouter.js
```diff
@@ -178,7 +182,12 @@ export default function AuthLayoutRouter (props) {
     return <Redirect to='/welcome' />
   }
 
-  if (currentGroupSlug && !currentGroup && !currentGroupLoading) {
+  if (!currentGroupMembership && hasDetail && paramPostId && currentGroupSlug) {
+    // this covers "Post I can see, and groupSlug for a group I cannot access." But it does not necessarily cover "Post I can see because its Public, but that isn't posted to any groups I have membership of"
+    return <Redirect push to={postUrl(paramPostId, { context: 'all', groupSlug: null })} />
+  }
+
+  if (currentGroupSlug && !currentGroup && !currentGroupLoading) { // this will have to be pushed down the hierarchy in the routers
```
tibetsprague, line 190: not sure what this comment is saying, didnt you just push it down already? is there more to do?

src/routes/AuthLayoutRouter/AuthLayoutRouter.js
```diff
@@ -178,7 +182,12 @@ export default function AuthLayoutRouter (props) {
     return <Redirect to='/welcome' />
   }
 
-  if (currentGroupSlug && !currentGroup && !currentGroupLoading) {
+  if (!currentGroupMembership && hasDetail && paramPostId && currentGroupSlug) {
+    // this covers "Post I can see, and groupSlug for a group I cannot access." But it does not necessarily cover "Post I can see because its Public, but that isn't posted to any groups I have membership of"
```
tibetsprague, line 186: lets have a comment that describes better what this is actually doing and why

src/util/navigation.js
```diff
@@ -92,9 +92,9 @@ export function groupDetailUrl (slug, opts = {}, querystringParams = {}) {
 export function postUrl (id, opts = {}, querystringParams = {}) {
   const action = get('action', opts)
   let result = baseUrl(opts)
-
   result = `${result}/post/${id}`
   if (action) result = `${result}/${action}`
+  console.log(result, 'ayayya', opts)
```
tibetsprague, line 97: remove console

tibetsprague (commented):
so to be clear this fixes half of part 1 of the 3 parts in that ticket right?

src/routes/AuthLayoutRouter/AuthLayoutRouter.js
thomasgwatson: I tried several ways of doing this and it would match only on one of the lines, and thus either give me the postId or the groupSlug, and as a result was causing other parts of the component to error

thomasgwatson:
> so to be clear this fixes half of part 1 of the 3 parts in that ticket right?

It effectively covers the first two of this set of three


1)    When navigating to a post path contextualized to a group the user doesn't have access to, but which is posted to a group they do have access to, they are currently shown an error page:
        They should instead be redirected to viewing the post in a context in which they are allowed to see the post. If the post is public, redirect to the Public context, or to the first group the post is in which the user does have access to (OR just to the All Groups context default).

2)    When navigating to a post path contextualized to a group (or Public) which the post is not part of, currently the post will render there:
        It should instead redirect according to the logic in item 1.

3)    Navigating to a a context-free post path (e.g. /post/12939) currently always redirects to the All Groups context (e.g. /all/post/12939).
        It should instead redirect to /public/post/12939 if the post is Public, otherwise branch to the recommended redirect logic in item 1.

src/routes/AuthLayoutRouter/AuthLayoutRouter.js
thomasgwatson: No, figured that there isn't any merit in pushing it down the component tree but missed removing comment, soz

thomasgwatson:
@tibetsprague this is ready to merge

tibetsprague (approved):

PR #1421, opened 2023-01-14
Make current user mentions be yellow orange

lorenjohnson:
Requires this to work, but won't break anything if release without it: 

https://github.com/Hylozoic/hylo-node/pull/916

### DIFF
--- a/src/components/PostListRow/PostListRow.scss
+++ b/src/components/PostListRow/PostListRow.scss
@@ -44,7 +44,7 @@ div .post-row:nth-child(1) {
 
 .type-author {
   display: flex;
-  align-items: end;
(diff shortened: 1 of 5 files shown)


src/components/PostListRow/PostListRow.scss
```diff
@@ -44,7 +44,7 @@ div .post-row:nth-child(1) {
 
 .type-author {
   display: flex;
-  align-items: end;
+  align-items: flex-end;
```
lorenjohnson, line 47: This was a listing thing. `end` should be the same as `flex-end` in this case and `end` is marginally supported...

thomasgwatson (approved):
lgtm

PR #1424, opened 2023-01-16
Fix several public facing routes

### LINKED ISSUE #1422: Going to public post using group URL should show the post
Going to a public post when not logged in using the group URL (e.g. https://www.hylo.com/groups/openteam/post/55661) should redirect to https://www.hylo.com/post/55661 and show the post, but instead it redirects to the public map

thomasgwatson:
There were a couple of public routes that were either inaccessible or their styling wasn't displaying correctly.

Closes #1422

### DIFF
--- a/src/routes/PublicLayoutRouter/PublicLayoutRouter.js
+++ b/src/routes/PublicLayoutRouter/PublicLayoutRouter.js
@@ -28,9 +28,10 @@ export default function PublicLayoutRouter (props) {
         <Route path={`/${POST_DETAIL_MATCH}`} exact component={PublicPostDetail} />
         <Route path='/:context(groups)/:groupSlug' exact component={PublicGroupDetail} />
         <Route path='/:context(public)/:view(map)' component={MapExplorerLayoutRouter} />
-        <Route path='/:context(public)/:view(groups)' exact component={GroupExplorerLayoutRouter} />
+        <Route path='/:context(public)/:view(groups)(.*)' exact component={GroupExplorerLayoutRouter} />
         {/* Remove this once we show the public stream */}
         <Redirect exact from={`/public/${POST_DETAIL_MATCH}`} to='/post/:postId' />
+        <Redirect exact from={`/:context(groups)/:groupSlug/:view(events|groups|map|members|projects|settings|stream|topics|custom)/${POST_DETAIL_MATCH}`} to='/post/:postId' />
         <Redirect to={{ pathname: '/public/map', state: { from: location } }} />
       </Switch>
       <HyloCookieConsent />
@@ -98,7 +99,9 @@ export function PublicPostDetail (props) {
 
   return (
     <div styleName='center-column' id={DETAIL_COLUMN_ID}>
+      <div />
       <PostDetail {...props} />
+      <div />
     </div>
   )
 }
@@ -133,7 +136,9 @@ export function GroupExplorerLayoutRouter () {
   return (
     <>
       <div styleName='center-column' id={CENTER_COLUMN_ID}>
-        <GroupExplorer />
+        <div>
+          <GroupExplorer />
+        </div>
       </div>
       <Route
         path={`(.*)/${GROUP_DETAIL_MATCH}`}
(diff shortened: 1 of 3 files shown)


src/routes/PublicLayoutRouter/PublicLayoutRouter.js
```diff
@@ -28,9 +28,10 @@ export default function PublicLayoutRouter (props) {
         <Route path={`/${POST_DETAIL_MATCH}`} exact component={PublicPostDetail} />
         <Route path='/:context(groups)/:groupSlug' exact component={PublicGroupDetail} />
         <Route path='/:context(public)/:view(map)' component={MapExplorerLayoutRouter} />
-        <Route path='/:context(public)/:view(groups)' exact component={GroupExplorerLayoutRouter} />
+        <Route path='/:context(public)/:view(groups)(.*)' exact component={GroupExplorerLayoutRouter} />
```
lorenjohnson, line 31: Did you try removing `exact`? I think that would do the same thing and be more idiomatic.

src/routes/PublicLayoutRouter/PublicLayoutRouter.js
```diff
@@ -28,9 +28,10 @@ export default function PublicLayoutRouter (props) {
...
         {/* Remove this once we show the public stream */}
         <Redirect exact from={`/public/${POST_DETAIL_MATCH}`} to='/post/:postId' />
+        <Redirect exact from={`/:context(groups)/:groupSlug/:view(events|groups|map|members|projects|settings|stream|topics|custom)/${POST_DETAIL_MATCH}`} to='/post/:postId' />
```
lorenjohnson, line 34: It may be a small rabbithole, but one I want us to keep an eye one as this I think is the case: This I think can be just `(.*)/${POST_DETAIL_MATCH}` and not break anything. Try it if you don't mind, and I am sorry ahead of time if I've missed something.

src/routes/PublicLayoutRouter/PublicLayoutRouter.js
```diff
@@ -133,7 +136,9 @@ export function GroupExplorerLayoutRouter () {
   return (
     <>
       <div styleName='center-column' id={CENTER_COLUMN_ID}>
-        <GroupExplorer />
+        <div>
```
lorenjohnson, line 139: Curious why the extra `div` fixed things here? It's nitpick a bit, but not a fan of extra containers unless they're truly necessary, in most all cases I find that adjusting the flex box layout of the existing container is good enough?

lorenjohnson (changes requested):
Don't stress it, but would be nice to hear back on these few things. And glad you're getting to this stuff!

src/routes/PublicLayoutRouter/PublicLayoutRouter.js
thomasgwatson: The whole complexity of the auth styling for this, with the nav div, the center div and the detail div, have not been ported over. This was an extra fix, beyond the scope of this ticket, just to have things looking vaguely correct.

lorenjohnson (approved):
I haven't QA'd through to test the routes, assume you did, and the code looks good👍🏼

src/routes/PublicLayoutRouter/PublicLayoutRouter.js
```diff
@@ -28,9 +28,8 @@ export default function PublicLayoutRouter (props) {
         <Route path={`/${POST_DETAIL_MATCH}`} exact component={PublicPostDetail} />
         <Route path='/:context(groups)/:groupSlug' exact component={PublicGroupDetail} />
         <Route path='/:context(public)/:view(map)' component={MapExplorerLayoutRouter} />
-        <Route path='/:context(public)/:view(groups)' exact component={GroupExplorerLayoutRouter} />
-        {/* Remove this once we show the public stream */}
-        <Redirect exact from={`/public/${POST_DETAIL_MATCH}`} to='/post/:postId' />
+        <Route path='/:context(public)/:view(groups)' component={GroupExplorerLayoutRouter} />
+        <Redirect from={`(.*)/${POST_DETAIL_MATCH}`} to='/post/:postId' />
```
tibetsprague, line 32: won't this break trying to look at a post from the public map? it will redirect to /post/id

src/routes/PublicLayoutRouter/PublicLayoutRouter.js
tibetsprague: won't this make it not possible to view posts from the public map though? /public/map/post/:id will redirect to /post/:id

src/routes/PublicLayoutRouter/PublicLayoutRouter.js
thomasgwatson: any public/map/post/:postId routes will be caught and forwarded to the 'MapExplorerLayoutRouter' before this redirect is invoked. AND... the mapExplorer is wonky after the routing shifts, which I am now addressing in the context of the map not rendering.

tibetsprague: ok cool

PR #1425, opened 2023-01-17
Public map displays to non-logged in users

### LINKED ISSUE #1366: Public map rendering seems borked
Likely this is some weirdness with how we are setting up mapboxGL....

https://www.hylo.com/public/map/?sortBy=updated for logged in user

Image

The same url for non-logged in user, incogntio

Image

Same url for non-logged in user, other browser

Image

The 'map' or WebGL warnings here might provide some clue to sorting this out

Image

thomasgwatson:
closes #1366 

Found another bug on closing detail pages in the non-logged-in public map, while working on this ticket. Will file a ticket.

### DIFF
--- a/src/routes/PublicLayoutRouter/PublicLayoutRouter.scss
+++ b/src/routes/PublicLayoutRouter/PublicLayoutRouter.scss
@@ -67,14 +67,20 @@
 }
 
 .center-column {
-  padding-bottom: 100px;
   display: flex;
   align-items: center;
   justify-content: center;
 }
 
-.center-column > div {
+.non-map-view {
+  padding-bottom: 100px;
+}
+
+.non-map-view > {
   margin-top: 40px;
+}
+
+.center-column > div {
   box-shadow: 0px 4px 15px rgba(42, 64, 89, 0.29);
   border-radius: 5px;
   background-color: rgba(255, 255, 255, 1);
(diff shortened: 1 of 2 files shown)


tibetsprague:
What does this ticket do? we want the map to display to non logged in users if they go to /public/map

tibetsprague:
oh I see, it fixes display of the map

src/routes/PublicLayoutRouter/PublicLayoutRouter.scss
```diff
@@ -67,14 +67,20 @@
...
   align-items: center;
   justify-content: center;
 }
 
-.center-column > div {
+.non-map-view {
+  padding-bottom: 100px;
+}
+
+.non-map-view > {
```
tibetsprague, line 79: do we really need both a .map-view and a .non-map-view class? seems like you could just use one or the other and have a default that then gets overridden by the one class being set?

thomasgwatson: I tried doing that; its effectively the center-column styling that is the default. However, there are some other stylings that depend on the center-column class being present, so I couldn't just replace it without replicating all of those as well. This ended up being the least amount of code change

tibetsprague (approved):

PR #1432, opened 2023-01-18
Evo 1430 my home banner

### LINKED ISSUE #1430: Add Header for My Home stream and fix top banner text
Also reordering the orrrrder!

- [x] change order
- [x] Add header, with user's banner
- [x] Change topNav copy to "Personal, My Home" instead of "Global, blah blah"

thomasgwatson:
Closes #1430


src/components/StreamBanner/StreamBanner.js
```diff
@@ -40,79 +42,85 @@ export default function StreamBanner ({
     avatarUrl = publicGlobe
     bannerUrl = allGroupsBanner
     // TODO list count of public posts and public groups in subtitle
-    subtitle = `All Posts Marked Public`
+    subtitle = 'All Posts Marked Public'
+  } else if (context === 'my') {
+    name = `My Home: ${capitalize(view)}`
+    avatarUrl = currentUser.avatarUrl || publicGlobe
```
tibetsprague, line 48: seems like we could have a better default than the globe, how about a house icon?

src/routes/AuthLayoutRouter/components/TopNav/TopNav.js
```diff
@@ -105,13 +107,15 @@ function Logo ({ group, isPublic }) {
   )
 }
 
-function Title ({ group, isPublic, onClick }) {
+function Title ({ group, isPublic, onClick, isMyHome }) {
   let [label, name] = ['GLOBAL', 'All My Groups']
```
tibetsprague, line 111: lets change GLOBAL here to PERSONAL

src/routes/AuthLayoutRouter/components/TopNav/TopNav.js
```diff
@@ -105,13 +107,15 @@ function Logo ({ group, isPublic }) {
...
 
   if (group) {
     [label, name] = [group.typeDescriptor, group.name]
   } else if (isPublic) {
     [label, name] = ['GLOBAL', 'Public Groups & Posts']
+  } else if (isMyHome) {
+    [label, name] = ['Personal', 'My Home']
```
tibetsprague, line 118: personal should probably be capitalized like GLOBAL is?

thomasgwatson: The styling automatically does this but I'm make it consistent in the copy (since we'll need to add more for i18n changes as well)

src/components/StreamBanner/StreamBanner.js
thomasgwatson: To fulfill this, we'd need a specific home image in '/public'. There is likely an icon in the font set, but the rendering component is expecting an image, not an icon-class, so it would also need to be edited and given an alternative styling to conditionally use the Icon component.

Do you want to escalate to Aaron to produce a Home graphic? Note, this backup will only render when a user does not have an avatarUrl (do they get a default one?)

tibetsprague: users do get a default actually, so maybe this isnt an issue.

tibetsprague (approved):

PR #1456, opened 2023-02-13
Add groupId to Mixpanel events and track more events

tibetsprague:
Add new POST_REACTION and COMMENT_REACTION and EVENT_RSVP events

Track the groups that a user is part of

### DIFF
--- a/src/routes/WelcomeWizardRouter/AddLocation/AddLocation.js
+++ b/src/routes/WelcomeWizardRouter/AddLocation/AddLocation.js
@@ -1,4 +1,5 @@
 import cx from 'classnames'
+import { AnalyticsEvents } from 'hylo-shared'
 import React, { Component } from 'react'
 import LocationInput from 'components/LocationInput'
 import { ensureLocationIdIfCoordinate } from 'components/LocationInput/LocationInput.store'
@@ -46,7 +47,7 @@ export default class AddLocation extends Component {
 
     this.props.updateUserSettings(changes)
       .then(() => {
-        this.props.trackAnalyticsEvent('Signup Complete')
+        this.props.trackAnalyticsEvent(AnalyticsEvents.SIGNUP_COMPLETE)
         this.props.goToNextStep()
       })
   }
--- a/src/routes/WelcomeWizardRouter/AddLocation/AddLocation.js
+++ b/src/routes/WelcomeWizardRouter/AddLocation/AddLocation.js
@@ -47,7 +47,7 @@ export default class AddLocation extends Component {
 
     this.props.updateUserSettings(changes)
       .then(() => {
-        this.props.trackAnalyticsEvent(AnalyticsEvents.SIGNUP_COMPLETE)
+        this.props.trackAnalyticsEvent(AnalyticsEvents.SIGNUP_PROFILE_FINISHED)
         this.props.goToNextStep()
       })
   }
--- a/src/routes/WelcomeWizardRouter/AddLocation/AddLocation.js
+++ b/src/routes/WelcomeWizardRouter/AddLocation/AddLocation.js
@@ -47,7 +47,7 @@ export default class AddLocation extends Component {
 
     this.props.updateUserSettings(changes)
       .then(() => {
-        this.props.trackAnalyticsEvent(AnalyticsEvents.SIGNUP_PROFILE_FINISHED)
+        this.props.trackAnalyticsEvent(AnalyticsEvents.SIGNUP_COMPLETE)
         this.props.goToNextStep()
       })
   }
--- a/src/store/actions/createComment.js
+++ b/src/store/actions/createComment.js
@@ -7,9 +7,10 @@ export default function createComment ({
   attachments,
   groupIds,
   parentCommentId,
-  postId,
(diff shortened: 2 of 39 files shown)


src/routes/WelcomeWizardRouter/AddLocation/AddLocation.js
```diff
@@ -46,7 +47,7 @@ export default class AddLocation extends Component {
 
     this.props.updateUserSettings(changes)
       .then(() => {
-        this.props.trackAnalyticsEvent('Signup Complete')
```
lorenjohnson, line 49: Without a careful analysis, I wonder if "Signup Profile Finished" (the value of `SIGNUP_PROFILE_FINISHED`) vs the previous  "Signup Complete" will create some discontinuity in analytics data previous to this change. I do see that we still send "Signup Complete" later now.

src/store/actions/createComment.js
```diff
@@ -27,7 +29,14 @@ export default function createComment ({
       postId,
       text,
       attachments,
-      analytics: AnalyticsEvents.COMMENT_CREATED
+      analytics: {
+        eventName: AnalyticsEvents.COMMENT_CREATED,
+        commentLength: TextHelpers.textLengthHTML(text),
+        groupId: post.groups.map(g => g.id),
```
lorenjohnson, line 35: Just curious why you called it `groupId` vs `groupIds` and always require an array (which may be the case).

lorenjohnson (approved):
Other than comments, I think this LGTM 👍🏼

src/store/actions/createComment.js
tibetsprague: Thats how Mixpanel wanted it, they wanted a group Key and suggested it be singular, but then you can pass in an array for multiple values

src/routes/WelcomeWizardRouter/AddLocation/AddLocation.js
tibetsprague: I mean it's slightly different, now we call Signup Complete when they have registered, instead of waiting until they have filled in Profile photo and location (when we call SIGNUP_PROFILE__FINISHED. So yes the same event will mean something slightly different now, but I think thats ok. Moving forward this is more accurate.

tibetsprague: we actually send Signup Complete sooner now

lorenjohnson: I know this is a bit confusing, and hard to track but within the Web and the similarly in the App the `AuthState` object reflects as `AuthState.SignupInProgress` (as `user.signupInProgress` is still true) where you're currently sending  `AnalyticsEvents.SIGNUP_COMPLETE`. This `signupInProgress` indicator means the user will be  redirected back into the signup process upon login if they leave and return. 

* I think we should be sending `AnalyticsEvents.SIGNUP_COMPLETE` only once `signupInProgress` is set to false to keep things consistent between these two worlds. 

* Where you're currently sending `AnalyticsEvents.SIGNUP_COMPLETE` I would create an event called `AnalyticsEvents.REGISTRATION_COMPLETE`. 

* Consider removing `SIGNUP_` from `SIGNUP_EMAIL_VERIFICATION_SENT` and `SIGNUP_EMAIL_VERIFIED`?

lorenjohnson: Sorry to complicate this at all, but if you follow and agree, I think this is one of those spots where we can avoid creating accidental complexity / unnecessary layers or indirection where we may thank ourselves later.

tibetsprague: I'm fine with all that

tibetsprague: though I do kind of like having SIGNUP_ in front of all 4 of those events because they are all part of the signup funnel, not attached though

tibetsprague: so then SIGNUP_REGISTRATION_COMPLETE and SIGNUP_COMPLETE

lorenjohnson: Cool with that.

tibetsprague: though, the problem with this is it is harder for someone to look at and understand, we'll have to explain to people. this is why I switched it to SIGNUP_PROFILE_FINISHED 🤷‍♂️

tibetsprague: though, the problem with this is it is harder for someone to look at and understand, we'll have to explain to people. this is why I switched it to SIGNUP_PROFILE_FINISHED 🤷‍♂️

lorenjohnson: Yeah, SIGNUP_COMPLETE means exactly that in the ÅuthState world, and I would think we should keep that event name for the final step.

PR #1468, opened 2023-03-13
Fix some routing issues when not logged in

tibetsprague:
When viewing a /groups... page either show the public groups page for public groups or redirect to login, instead of redirecting to the public map

When redirecting to /login for non public group or post make sure to include correct returnToUrl


thomasgwatson (approved):
lgtm

PR #1491, opened 2023-04-11
1367-duplicate-topics

tibetsprague:
Fixed duplicate topics appearing in topic selector. And fix order to show more popular ones at the top


thomasgwatson (approved):
lgtm

PR #1497, opened 2023-04-18
Return close button to non-logged-in public map

### LINKED ISSUE #1426: Closing Detail panel on public map does not work for non-logged-in users
Reproduce:
- open `/public/maps` in a non-logged-in browser context
- click on a group or post (opens respective Detail panel)
- click on the X icon on top right of screen to close group or post detail panel
- goes to login page...because the onClick function has not got access to RouteParams, and so feeds bad inputs to `baseUrl` function

thomasgwatson:
closes #1426


src/routes/PublicLayoutRouter/PublicLayoutRouter.scss
```diff
@@ -16,6 +16,13 @@
   }
 }
 
+.wrapper {
```
tibetsprague, line 19: this seems like a significant enough thing to require some more testing for all use cases, i will do that

src/components/PostCard/PostHeader/PostHeader.js
```diff
@@ -156,7 +156,7 @@ export default class PostHeader extends PureComponent {
               {type && <PostLabel type={type} styleName='label' />}
               {dropdownItems.length > 0 &&
                 <Dropdown toggleChildren={<Icon name='More' />} items={dropdownItems} alignRight />}
-              {close && context &&
```
tibetsprague, line 159: any idea why this was here?

src/routes/PublicLayoutRouter/PublicLayoutRouter.scss
thomasgwatson: Yep. This is returning the public layout to more similar to the auth layout, since those diverged with routing changes. There are still some differences styling things that are a little off but didn't want to get any more sidetracked from the scope of the ticket that this.

tibetsprague:
ok in testing i found an issue - if you go to /public/map or /public/groups and open a group and then close it, it takes you to the login screen instead of back to the map or explorer

tibetsprague (approved):

PR #1518, opened 2023-06-07
Some quick fixes

tibetsprague:
Fix a few translation related issues
Discusion with only 1 's' also was breaking create modal somewhat

Fix display of member posts on user profile

Fixes to Group Details page:

- Always show privacy settings in main body, instead of ugly random text describing accessibility setting
- Hide Posts and members in body for now since we never show anything there
- Move member count to header
- Fix closing about info from Explore page
- Add header text asking people to asnwer join questions

Fixes https://github.com/Hylozoic/hylo-evo/issues/1493


thomasgwatson (approved):
lgtm

PR #1555, opened 2023-10-11
Let chatroom scroll to a specific chat

thomasgwatson:
Contributes to https://github.com/Hylozoic/HyloReactNative/issues/663

This should be SAFE to merge, it only adds new functionality, no changes to routing etc

### DIFF
--- a/src/routes/ChatRoom/ChatRoom.connector.js
+++ b/src/routes/ChatRoom/ChatRoom.connector.js
@@ -84,7 +84,7 @@ export function mapStateToProps (state, props) {
   const context = getRouteParam('context', state, props)
   const view = getRouteParam('view', state, props)
 
-  const querystringParams = getQuerystringParam(['search'], null, props)
+  const querystringParams = getQuerystringParam(['search', 'postid'], null, props)
   const search = getQuerystringParam('search', state, props)
 
   const imageAttachments = getAttachments(state, { type: 'post', id: 'new', attachmentType: 'image' })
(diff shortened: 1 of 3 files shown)


src/routes/ChatRoom/ChatRoom.js
```diff
@@ -371,6 +375,34 @@ export default function ChatRoom (props) {
     }, [])
   }, [postsPast, postsFuture])
 
+  useEffect(() => {
```
tibetsprague, line 378: Hmm. Seems like there should be a way to load the posts starting at the postid instead of loading and then scrolling to a new post

src/routes/ChatRoom/ChatRoom.connector.js
tibetsprague, line 1: Where the cursor is set in fetchPostParams can’t you just change the cursor to the postid we want to start at?

thomasgwatson: Yep, makes sense, I've changed it to that

src/routes/ChatRoom/ChatRoom.connector.js
```diff
@@ -84,8 +84,9 @@ export function mapStateToProps (state, props) {
   const context = getRouteParam('context', state, props)
   const view = getRouteParam('view', state, props)
 
-  const querystringParams = getQuerystringParam(['search'], null, props)
+  const querystringParams = getQuerystringParam(['search', 'postid'], null, props)
   const search = getQuerystringParam('search', state, props)
+  const postIdToStartAt = querystringParams?.postid
```
tibetsprague, line 89: Oh one more little thing, let’s call this postId

tibetsprague (approved):

thomasgwatson:
Yeah, I was getting it in my head that urls are case-insensitive but that
doesn't apply to query strings

Issue #1520, opened 2023-07-05
Change title field placeholder to "Add a title * "


tibetsprague:
Maybe add a tooltip to disabled button saying what is missing

Issue #923, opened 2021-08-06
Make it clear that people need to add a Title to a post


tibetsprague:
It does not show that the title is required or show an error message when not set, this has caused confusion for many

PR #1562, opened 2023-11-02
Fix post-call-to-action copy

### LINKED ISSUE #680: Change post title placeholder - "What's on your mind is not clear"
People dont know they need to fill it out. Also need to show an error message if it is not filled out.

Also we shouldnt use the same text in the comment box too.

thomasgwatson:
closes #680 #923 #1520


public/locales/en.json
```diff
@@ -768,8 +769,8 @@
   "What would you like to call your project?": "What would you like to call your project?",
   "What would you like to create?": "What would you like to create?",
   "What you will do at your next event?": "What you will do at your next event?",
-  "Whats on your mind?": "What's on your mind?",
-  "What’s on your mind?": "What’s on your mind?",
+  "Whats on your mind?": "Create a post",
```
tibetsprague, line 772: hmm. why do we have this one with broken grammar?

public/locales/en.json
```diff
@@ -768,8 +769,8 @@
...
+  "What’s on your mind?": "Create a post",
```
tibetsprague, line 773: Seems like we may as well change the key as well

src/components/PostEditor/PostEditor.js
```diff
@@ -185,12 +186,12 @@ class PostEditor extends React.Component {
     const { t } = this.props
 
     const titlePlaceHolders = {
-      offer: t('What help can you offer?'),
-      request: t('What are you looking for help with?'),
-      resource: t('What resource is available?'),
-      project: t('What would you like to call your project?'),
-      event: t('What is your event called?'),
-      default: t('What’s on your mind?')
+      offer: t('Add title'),
```
tibetsprague, line 189: maybe you discusses as a group but I think "Add a title" is clearer

tibetsprague (approved):

tibetsprague:
oops, looking at this again and noticing that you say that a post requires a Description, which it doesn't, only a Title actually

Issue #1556, opened 2023-10-18
UI is off on Edit Profile when linking social accounts 


KrishaTerranIO:
UI element blocking social links content in the 'edit profile menu'.

<img width="727" alt="image" src="https://github.com/Hylozoic/hylo-evo/assets/92816219/221e2f34-0810-431e-bf77-6f53637cf67c">

Issue #1554, opened 2023-10-09
Instructional message when adding images on comments


KrishaTerranIO:
Adding instructional text when adding an image as a comment on a post that text is needed

KrishaTerranIO:
@thomasgwatson

Issue #1553, opened 2023-10-09
Instructional message about post title


KrishaTerranIO:
Posts Title should have instructions that Titles are required in order to post

KrishaTerranIO:
@thomasgwatson

tibetsprague:
still wondering where this one actually is addressed

thomasgwatson:
In:
https://github.com/Hylozoic/hylo-evo/pull/1562

Or do you want a little more? The title placeholder is changed to something
clearer and the post button has a tooltip has an explainer

tibetsprague:
Ah, ok that should be enough

PR #1563, opened 2023-11-06
Added various instructional messages and small fixes

### LINKED ISSUE #1552: Instructional message when adding topics
When folks try to add topics with a space between (Ex: #product design), there should be a popup or instructions that there cannot be a space between (or emphasizing to include a hyphen). If folks try to add a topic with a space, there is no way for them to create the post

thomasgwatson:
Closes #1552 #1553 #1554 #1556 

Screenshot 2023-11-06 at 8 51 09 AM
Screenshot 2023-11-06 at 8 50 33 AM
Screenshot 2023-11-06 at 8 50 07 AM

### DIFF
--- a/src/components/HyloEditor/HyloEditor.js
+++ b/src/components/HyloEditor/HyloEditor.js
@@ -149,7 +149,7 @@ const HyloEditor = React.forwardRef(function HyloEditor ({
       return editorRef.current.getText()
     },
     isEmpty: () => {
-      return editorRef.current?.isEmpty
+      return editorRef.current ? editorRef.current.isEmpty : true
     },
     setContent: content => {
       editorRef.current.commands.setContent(content)
--- a/src/components/HyloEditor/HyloEditor.js
+++ b/src/components/HyloEditor/HyloEditor.js
@@ -149,7 +149,7 @@ const HyloEditor = React.forwardRef(function HyloEditor ({
       return editorRef.current.getText()
     },
     isEmpty: () => {
-      return editorRef.current.isEmpty
+      return editorRef.current?.isEmpty
     },
     setContent: content => {
       editorRef.current.commands.setContent(content)
--- a/src/routes/PostDetail/Comments/CommentForm/CommentForm.js
+++ b/src/routes/PostDetail/Comments/CommentForm/CommentForm.js
@@ -11,6 +11,7 @@ import HyloEditor from 'components/HyloEditor'
 import Icon from 'components/Icon'
 import Loading from 'components/Loading'
 import RoundImage from 'components/RoundImage'
+import Tooltip from 'components/Tooltip'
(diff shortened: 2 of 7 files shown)


src/components/HyloEditor/HyloEditor.js
```diff
@@ -149,7 +149,7 @@ const HyloEditor = React.forwardRef(function HyloEditor ({
       return editorRef.current.getText()
     },
     isEmpty: () => {
-      return editorRef.current.isEmpty
+      return editorRef.current?.isEmpty
```
tibetsprague, line 152: wait, shouldnt editorRef.current being not existent return true? so in that case it would be editorRef.current ? editorRef.current.isEmpty : true

src/routes/PostDetail/Comments/CommentForm/CommentForm.js
```diff
@@ -51,11 +52,12 @@ class CommentForm extends Component {
...
       sendIsTyping,
       attachments,
-      clearAttachments
+      clearAttachments,
+      t
     } = this.props
 
     if (this.editor?.current && this.editor.current.isEmpty()) {
-      // Do nothing and stop event propagation
+      window.alert(t('You need to include text to post a comment'))
```
tibetsprague, line 60: you know, it seems like we should allow comments that are just images... but maybe thats another ticket. I could also see this one being kind of annoying. I dont think anyone would expect that hitting enter with an empty comment would do anything, but now it pops up a window alert you have to close. maybe we only do an alert if there's an image but no text? or maybe we just change it so that if there's an image but no text we set the comment text to a blank space and let it be created?

tibetsprague:
where's the part about telling people they must add a Title to a post?

thomasgwatson:
> where's the part about telling people they must add a Title to a post?

Actually covered in another PR.

src/components/HyloEditor/HyloEditor.js
thomasgwatson: With the other changes here, there is a moment in its lifecycle where `editorRef.current` is null or undefined when this.editor.current.isEmpty() is called, and so it errors.

src/routes/PostDetail/Comments/CommentForm/CommentForm.js
thomasgwatson: Ok, changed this to allow comment posts with just images.

src/components/HyloEditor/HyloEditor.js
tibetsprague: I get that, it just seems like for isEmpty to return the correct value (which should be true in that case) then it needs to be rewritten as I suggested

thomasgwatson: Ahh yes sorry, I think I just saw the first half of your comment in the email and missed the second half

PR #1579, opened 2024-01-19
792-join-questions-always

tibetsprague:
Enable join questions being asked on group invigte and for closed and open groups


thomasgwatson (approved):
lgtm

PR #1583, opened 2024-02-02
Evo 1299 increase title len

### LINKED ISSUE #1299: Allow for longer post titles - 80 characters
Make sure to test how posts look at all sizes of viewport in all different views

KevinTriplett:
closes #1299 increase title length from 50 to 80, with tests and updates to other files to allow passing tests.

In the following screenshots, ignore the comment, the full 80 character title is shown

Screenshot 2024-02-01 at 11 54 26 AM
Screenshot 2024-02-01 at 11 55 37 AM
Screenshot 2024-02-01 at 11 55 47 AM
Screenshot 2024-02-01 at 11 55 57 AM
Screenshot 2024-02-01 at 11 56 06 AM
Screenshot 2024-02-01 at 11 56 15 AM
Screenshot 2024-02-01 at 11 56 27 AM
Screenshot 2024-02-01 at 11 56 42 AM
Screenshot 2024-02-01 at 11 56 52 AM


src/components/PostEditor/PostEditor.js
```diff
@@ -249,7 +249,7 @@ class PostEditor extends React.Component {
 
   handleTitleChange = (event) => {
     const title = event.target.value
-    title.length >= MAX_TITLE_LENGTH
+    title.length > MAX_TITLE_LENGTH
```
tibetsprague, line 252: so we actually do want a warning to appear when you have reached maximum length. otherwise there would be no way to know you had reached the end. the current error is not great and could definitely be improved by making it a less intense notice that you have hit the max length, but also that would increase the scope of this ticket. For now I think we should leave the way it is

src/components/PostEditor/PostEditor.test.js
```diff
@@ -27,6 +27,7 @@ describe('PostEditor', () => {
   it('renders announcement option with admin in props', () => {
     const props = {
       ...baseProps
+      // KMT: where is admin?
```
tibetsprague, line 30: hmm, gooood question

src/components/PostEditor/PostEditor.test.js
```diff
@@ -136,6 +137,7 @@ describe('PostEditor', () => {
       setIsDirty: jest.fn()
     }
 
+    // KMT: these next two blocks look the same?
     test('form in editing mode', () => {
```
tibetsprague, line 141: weirdddd, it should be passing editing: true to the first one, want to try that?

tibetsprague:
There's a few more things/places I want to make sure you've tested 

- [ ] The List view of the Stream
- [ ] The Large Grid view of the Stream
- [ ] The Small Grid view of the stream
- [ ] The Explore page Upcoming Events, and other widgets
- [ ] Map page drawer at different sizes
- [ ] All of these places for a Request with a location, several tags, and a start time and end time set (to make sure all the elements appear without overlap in all the views/sizes of the app

src/components/PostEditor/PostEditor.js
KevinTriplett: tbh, it seems this test can't fail normally because the html input control is limited at `MAX_TITLE_LENGTH` -- I think the only way around it would be a hack.

src/components/PostEditor/PostEditor.test.js
KevinTriplett: I did this and tests pass:
````    test('form in editing mode', () => {
      const wrapper = shallow(<PostEditor {...props}  editing={true} />)
      expect(wrapper).toMatchSnapshot()
    })

KevinTriplett:
Thanks for advising testing it as an event and resource and such, with the different views. It seems well behaved, if clipping the title is what's desired / expected. Opening the item displays the full title with no overlapping. Looks good, but I wouldn't say I did an exhaustive test, more like a sample with no obvious errors.

src/components/PostEditor/PostEditor.test.js
KevinTriplett: Yeah, there's no `admin` option in the `PostEditor.js` file. Is it an external influence?

src/components/PostEditor/PostEditor.js
tibetsprague: im confused, I still think we want this to be >= so the error/warning message appears once you've reached the maximum. isn't this the place that does that?

tibetsprague:
looks good except for my one comment about the error message

src/components/PostEditor/PostEditor.js
KevinTriplett: Ahhhhh I see -- yeah, that's not an error but I understand that it needs to pretend it's an error in order to let the user know they hit the max. Hrm. I agree, out of scope, let's open a new ticket to soften that error. I'll undo this change and revise the test and push.

KevinTriplett: Okay, code and test updated

KevinTriplett:
Test and code reverted

src/components/PostEditor/PostEditor.test.js
tibetsprague: what is meant by admin here is actually canModerate. you could get the test setup correctly to show the editor as a group moderator and then it should show the announcement option.

tibetsprague (approved):
good enough for me!

KevinTriplett:
@tibetsprague I've scheduled a zoom meet for you to walk me through the way shallow wrappers work. I've coded something that seems like it should work but the wrapper `find` method is returning an empty object despite me asking for even just a `'div'` element. Reading the manuals isn't helping me for some reason. Thanks

PR #1589, opened 2024-02-15
Fix i18n for account creation

### LINKED ISSUE #1498: Integrate i18n into onboarding workflow
Locale isn't currently factored in via the signup process. There is no language selector, and so the emails and language will all be in EN for now. Need to carve out time to identify what needs to be done there.

thomasgwatson:
closes #1498 

Screenshot 2024-02-14 at 4 57 23 PM


tibetsprague (approved):
cool!

Issue #1244, opened 2022-07-06
Add new notification setting to get email and/or push notification for every post in a group


tibetsprague:
Several folks have requested a setting to get a new email and/or push notification for every post in a group, so it feels more like slack. Need to think through the UI of the notification settings for this

tibetsprague:
notes, ideas here: https://docs.google.com/document/d/1oLZfRv2E7ROGTHOTyUtGTf055wb-2pwWUybgchwasCQ/edit#heading=h.gmwbrughaqcy

PR #1591, opened 2024-02-16
Add setting to receive a notification for every post

tibetsprague:
Or just importantp posts = announcements and mentions

Fixes https://github.com/Hylozoic/hylo-evo/issues/1244


src/store/models/Notification.js
```diff
@@ -56,10 +57,13 @@ export function urlForNotification ({ activity: { action, post, comment, group,
...
     case ACTION_NEW_COMMENT:
       return postCommentUrl({ postId: post.id, commentId: comment.id, groupSlug })
-    case ACTION_TAG:
+    case ACTION_TAG: {
       const tagReason = find(r => r.startsWith('tag: '), reasons)
       const topicName = tagReason.split(': ')[1]
       return postUrl(post.id, { groupSlug, topicName })
+    }
+    case ACTION_NEW_POST:
+      return postUrl(post.id, { groupSlug })
```
thomasgwatson, line 66: This reminds me that it would be good to have a different path when the post is a chat-post, so it goes straight to the chat room. I thought I worked on something like that but might just have been chat paths from node repo

thomasgwatson (approved):
lgtm

src/store/models/Notification.js
tibetsprague: you are right, I will do this

tibetsprague: oh actually this already works, if there is a topic on the post it goes to ACTIOB_TAG above which goes to the chat room with the post open

thomasgwatson:
ahh yes, tags => topics

thomasgwatson:
Although... maybe it doesn't like it for #general ? I clicked on this notification (below pic) and it sent me to that url, which is the regular post one

Screenshot 2024-02-19 at 1 00 18 PM

thomasgwatson:
Yeah, it worked on an earlier Krisha chat notif. The mention must get picked up before the topic/tag

tibetsprague:
Ooh, I can move ACTION_TAG above mention to prioritize that URL

tibetsprague:
ok fixed so if someone is mentioned in a chat room is goes to the chat room

PR #1613, opened 2024-04-10
Fix for dup post change of type losing content

### LINKED ISSUE #1610: Changing post type of duplicated post before saving it loses all content
because changing the post type changes the URL and gets rid of the fromPostId= parameter

KevinTriplett:
Fixes #1610:

- Preserve `fromPostId` in search query param string for duplicated posts while editing
- Adds new function `setQuerystringParam` to support this preservation
- No new tests added (requiresintegration test?)

### DIFF
--- a/src/components/PostEditor/PostEditor.js
+++ b/src/components/PostEditor/PostEditor.js
@@ -27,6 +27,7 @@ import { PROJECT_CONTRIBUTIONS } from 'config/featureFlags'
 import { MAX_POST_TOPICS } from 'util/constants'
 import { sanitizeURL } from 'util/url'
 import Tooltip from 'components/Tooltip'
+import setQuerystringParam from 'store/selectors/setQuerystringParam'
 
 export const MAX_TITLE_LENGTH = 80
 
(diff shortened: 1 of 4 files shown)


src/components/PostEditor/PostEditor.js
```diff
@@ -27,6 +27,7 @@ import { PROJECT_CONTRIBUTIONS } from 'config/featureFlags'
 import { MAX_POST_TOPICS } from 'util/constants'
 import { sanitizeURL } from 'util/url'
 import Tooltip from 'components/Tooltip'
+import setQuerystringParam from 'store/selectors/setQuerystringParam'
```
tibetsprague, line 30: this is not a selector. selectors are just about getting data. we already have a function util/navigation.js#addQuerystringToPath which does almost this. you can tweak it to support no path, or add another function to util/navigation.js

tibetsprague (changes requested):

src/components/PostEditor/PostEditor.js
KevinTriplett: Doh! Thanks for pointing me to the right category :)

KevinTriplett: Okay, moved the function to util/navigation.js as addQuerystringToPath would have been kinda mangled by this feature.

tibetsprague (approved):
👍

PR #1618, opened 2024-04-12
Fix bug preventing image deletion on post

### LINKED ISSUE #1616: Unable to delete images in postEditor
The delete image `Icon` is a child in the dnd-kit `SortableContext` tree which prevents its click event from firing.

This interface looks promising: https://docs.dndkit.com/api-documentation/draggable/usedraggable

KevinTriplett:
fixes #1616 

Discovered that the 'X' delete icon was wrapped in the draggable image element. This prevented the click event from firing to delete the image.

After fixing the click event bug, I still could not delete a single image (it would delete all images) so I discovered that the `reject` method wasn't working properly with all the hash keys being passed in, so I isolated the `url` key and that fixed the problem. I didn't create any tests (other than updating the snapshot) but hand tested it with multiple images, sorting, and file attachments and it worked as expected.

### DIFF
--- a/src/components/AttachmentManager/AttachmentManager.js
+++ b/src/components/AttachmentManager/AttachmentManager.js
@@ -164,10 +164,9 @@ export function ImagePreview (props) {
   }
 
   return (
-    <div styleName='image-preview' ref={setNodeRef} style={style} {...listeners} {...attributes}>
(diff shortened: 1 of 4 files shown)


thomasgwatson (approved):
lgtm

PR #1619, opened 2024-04-16
Evo 1564 proposals v1

### LINKED ISSUE #1564: Proposal post type
Most up to date details here: https://docs.google.com/document/d/1EUiiTtIyGAIEZRwF1uJtDk-FAyj3HchRZ5Kwf62cSv8/edit#heading=h.j8sxm8kc5x8

https://www.figma.com/file/H3WNvkSDPvLsWKQqHIUHRc/Proposal-Post-Type?type=design&node-id=0%3A1&mode=design&t=jfMzRN7jEsDOEP5g-1

- [x] Ability to create proposal posts
- - [x] Set quorum
- - [x] Setup proposal options
- - [x] Templates/shortcuts to different voting options that we offer as defaults
- [ ] Display of proposals, including voting options and # of votes for each option. How do we want to display vote counts? Need designs.
- [ ] Ability to vote on them by choosing options, until the proposal closes. And you can leave a comment as you vote which shows up as a comment on the post that also includes what option the person voted on at the front of the comment somehow
- [ ] Ability for post creator to change the state of the proposal to discussion phase, voting phase, or complete. And outcome to passed or failed. 
- [ ] New Governance view that shows all current open proposals and also past proposals with whether the passed or failed or quorum not reached 
- [ ] Show on proposal whether quorum of votes needed for proposal to complete has been achieved or not (and how many more needed)
- [ ] Proposals cant be edited in voting phase

thomasgwatson:
closes #1564 

Screenshot 2024-04-15 at 1 37 48 PM
Screenshot 2024-04-11 at 4 46 47 PM
Screenshot 2024-04-11 at 4 46 58 PM
Screenshot 2024-04-11 at 4 47 13 PM
Screenshot 2024-04-11 at 4 47 24 PM
Screenshot 2024-04-11 at 4 47 34 PM
Screenshot 2024-04-11 at 4 47 43 PM


src/components/Map/layers/clusterLayer/index.js
```diff
@@ -133,7 +133,7 @@ class PostClusterLayer extends CompositeLayer {
               .map(f => f.properties)
 
             const sprite = ['event', 'member', 'offer', 'request', 'resource', 'project', 'discussion'].reduce((result, type) => {
-              return features.find(feature => feature.type === type) ? result.concat(type + 's') : result
+              return features.find(feature => feature.type === type || (type === 'discussion' && feature.type === 'proposal')) ? result.concat(type + 's') : result
```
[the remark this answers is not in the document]
tibetsprague: can you explain this? its using the discussion color for now? we did talk about making the color of request and offer the same (I think green) and then using yellow for proposals.

src/components/PostCard/PostBodyProposal/PostBodyProposal.js
```diff
@@ -0,0 +1,140 @@
...
+import {
+  addProposalVote,
+  removeProposalVote,
+  swapProposalVote
+} from '../../../store/actions/proposals'
+import QuorumBar from 'components/QuorumBar/QuorumBar'
+import './PostBodyProposal.scss'
+import RoundImageRow from 'components/RoundImageRow'
+
+const calcNumberOfVoters = (votes) => {
```
tibetsprague, line 17: it feels like these should be memoized values in the render that get recalculated when the votes change

src/components/PostCard/PostBodyProposal/PostBodyProposal.js
```diff
@@ -0,0 +1,140 @@
...
+  isAnonymousVote,
+  startTime,
+  quorum,
+  endTime,
+  groups,
+  id
+}) {
+  const dispatch = useDispatch()
+  const { t } = useTranslation()
+  const proposalOptionsArray = (proposalOptions && proposalOptions.items) ? proposalOptions?.items : []
```
tibetsprague, line 49: cant this just be `proposalOptions?.items || []` ?

src/components/PostCard/PostBodyProposal/PostBodyProposal.js
```diff
@@ -0,0 +1,140 @@
...
+  const proposalVotesArray = (proposalVotes && proposalVotes.items) ? proposalVotes?.items : []
+
+  const [currentUserVotes, setCurrentUserVotes] = useState(proposalVotesArray.filter(vote => vote?.user?.id === currentUser.id))
+  const [currentUserVotesOptionIds, setCurrentUserVotesOptionIds] = useState(currentUserVotes.map(vote => vote.optionId))
+  const [proposalVoterCount, setProposalVoterCount] = useState(calcNumberOfVoters(proposalVotesArray))
```
tibetsprague, line 54: yeah instead of having a useState and a useEffect I think you can just use useMemo for all these calculated values

src/components/PostCard/PostBodyProposal/PostBodyProposal.js
```diff
@@ -0,0 +1,140 @@
...
+  useEffect(() => {
+    setCurrentUserVotesOptionIds(currentUserVotes.map(vote => vote.optionId))
+  }, [currentUserVotes])
+
+  useEffect(() => {
+    setNumberOfPossibleVoters(calcNumberOfPossibleVoters(groups))
+  }, [groups])
+
+  function handleVote (optionId) {
+    if (proposalType === PROPOSAL_TYPE_SINGLE || proposalType === PROPOSAL_TYPE_MAJORITY) {
```
tibetsprague, line 71: question if we want to build in MAJORITY. why pick that one option when people might want 60% or 70%? thats why I think we should just leave it up to them to decide the outcome based on the votes

src/components/PostCard/PostBodyProposal/PostBodyProposal.js
```diff
@@ -0,0 +1,140 @@
...
+  return (
+    <div styleName={cx('proposal-body-container', { discussion: proposalStatus === PROPOSAL_STATUS_DISCUSSION, voting: proposalStatus === PROPOSAL_STATUS_VOTING, casual: proposalStatus === PROPOSAL_STATUS_CASUAL, completed: proposalStatus === PROPOSAL_STATUS_COMPLETED })}>
+      <div styleName={cx('proposal-status')}>
+        {proposalStatus === PROPOSAL_STATUS_DISCUSSION && t('Discussion in progress')}
+        {proposalStatus === PROPOSAL_STATUS_VOTING && t('Voting open')}
+        {proposalStatus === PROPOSAL_STATUS_COMPLETED && t('Voting ended')}
+        {proposalStatus === PROPOSAL_STATUS_CASUAL && t('Voting open')}
+      </div>
+      <div styleName={cx('proposal-timing')}>
+        {!startTime && t('Open timeframe')}
```
tibetsprague, line 101: to discuss: this language. Open timeframe is not totally clear to me but I dont immediately have a better idea

src/components/PostEditor/PostEditor.js
```diff
@@ -22,13 +22,18 @@ import DatePicker from 'components/DatePicker'
...
 import styles from './PostEditor.scss'
 import { PROJECT_CONTRIBUTIONS } from 'config/featureFlags'
 import { MAX_POST_TOPICS } from 'util/constants'
+import { setQuerystringParam } from 'util/navigation'
 import { sanitizeURL } from 'util/url'
 import Tooltip from 'components/Tooltip'
-import { setQuerystringParam } from 'util/navigation'
 
+
+const emojiOptions = ['', '1️⃣', '2️⃣', '3️⃣', '4️⃣', '5️⃣', '6️⃣', '7️⃣', '8️⃣', '9️⃣', '🔟', '✅✅', '👍', '👎', '⁉️', '‼️', '❓', '❗', '🚫', '➡️', '🛑', '✅', '🛑🛑', '🌈', '🔴', '🔵', '🟤', '🟣', '🟢', '🟡', '🟠', '⚫', '⚪', '🤷🤷', '📆', '🤔', '❤️', '👏', '🎉', '🔥', '🤣', '😢', '😡', '🤷', '💃🕺', '⛔', '🙏', '👀', '🙌', '💯', '🔗', '🚀', '💃', '🕺', '💯',]
```
tibetsprague, line 36: why limit emoji options? what if people want to choose between an Italian restaurant and a Chinese restaurant?

src/components/PostEditor/PostEditor.js
```diff
@@ -124,6 +134,7 @@ class PostEditor extends React.Component {
     t('Let people know about available resources')
     t('Create a project that people can help with')
     t('Invite people to your event')
+    t('Suggest a proposal for others to vote on')
```
tibetsprague, line 137: Maybe "Offer" instead of "Suggest"?

src/components/PostEditor/PostEditor.js
```diff
@@ -725,13 +804,162 @@ class PostEditor extends React.Component {
...
+          {isProposal && proposalOptions.length === 0 && (
+            <div styleName='footerSection'>
+              <div styleName='footerSection-label'>{t('Proposal template')}</div>
+
+              <div styleName='inputContainer'>
+                <Dropdown
+                  styleName='dropdown'
+                  toggleChildren={
+                    <span styleName='dropdown-label'>
+                      {t('Select pre-set')}
```
tibetsprague, line 816: to discuss: "Select one of our suggested voting templates" ?

src/components/PostEditor/PostEditor.js
```diff
@@ -725,13 +804,162 @@ class PostEditor extends React.Component {
...
+                  styleName='dropdown'
+                  toggleChildren={
+                    <span styleName='dropdown-label'>
+                      {proposalType ? t(`${proposalType}`) : t('Select pre-set')}
+                      <Icon name='ArrowDown' blue />
+                    </span>
+                  }
+                  items={[
+                    { label: t('Single Vote per person'), onClick: () => this.handleSetProposalType(PROPOSAL_TYPE_SINGLE) },
+                    { label: t('Multi-vote, unrestricted'), onClick: () => this.handleSetProposalType(PROPOSAL_TYPE_MULTI_UNRESTRICTED) },
```
tibetsprague, line 931: To discuss: this text, maybe "Voters can choose more than one option"

src/components/PostEditor/PostEditor.js
```diff
@@ -725,13 +804,162 @@ class PostEditor extends React.Component {
...
+                    { label: t('Majority'), onClick: () => this.handleSetProposalType(PROPOSAL_TYPE_MAJORITY) }
```
tibetsprague, line 932: probably remove

src/components/PostEditor/PostEditor.js
```diff
@@ -725,13 +804,162 @@ class PostEditor extends React.Component {
...
+              <SliderInput percentage={post.quorum} setPercentage={this.handleSetQuorum} />
+            </div>
+          )}
+          {isProposal && (
+            <AnonymousVoteToggle
+              isAnonymousVote={!!post.isAnonymousVote}
+              toggleAnonymousVote={this.toggleAnonymousVote}
+            />
+          )}
+          {/* {isProposal && (
```
tibetsprague, line 950: to discuss: can casual or strict be toggled just by whether there is a timeline or not? then we wouldn't need another control. would need to explain well though

src/components/PostEditor/PostEditor.js
```diff
@@ -851,7 +1079,15 @@ class PostEditor extends React.Component {
             valid={valid}
             loading={loading}
             submitButtonLabel={this.buttonLabel()}
-            save={() => this.save()}
+            save={() => {
+              if (isProposal && !isEqual(proposalOptions, this.props.post.proposalOptions?.items)) {
+                if (window.confirm(t('Changing proposal options will reset the votes. Are you sure you want to continue?'))) {
```
tibetsprague, line 1084: To discuss: if this happens we would probably need to send a notification to every person who voted, something to think about adding to the back end

src/components/Widget/FarmMapWidget/FarmMap.scss
```diff
@@ -34,6 +34,10 @@
   color: $color-discussions;
 }
 
+.postTip.proposal {
+  color: $color-discussions;
```
tibetsprague, line 38: or make the yellow color

src/graphql/fragments/postFieldsFragment.js
```diff
@@ -125,7 +126,31 @@ const postFieldsFragment = withComments => `
     region
   }
   peopleReactedTotal
-  myVote
+  proposalStatus
+  proposalType
+  quorum
+  proposalOptions {
```
tibetsprague, line 132: so here is where if we are always loading this for every post the backend must only actual do this query for proposals

tibetsprague (changes requested):
so goooood

src/components/Map/layers/clusterLayer/index.js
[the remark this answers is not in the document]
thomasgwatson: https://www.hylo.com/my/posts/post/72248

We didn't have a conclusion to that particular discussion

src/components/PostCard/PostBodyProposal/PostBodyProposal.js
thomasgwatson: @cpolitano I think this is one of the copy things Tibet was referring to a few sessions ago

src/components/PostEditor/PostEditor.js
thomasgwatson: @cpolitano copy questions

src/components/PostEditor/PostEditor.js
thomasgwatson: 1) In the balance between max flexibility and channeling users to the most useful/sensible options, channeling makes more sense for proposals.

2) Can't do double emojis with the regular emoji picker. (Although I note you aren't a fan of double emojis but they are a clear way to handle gradients of agreement)

src/graphql/fragments/postFieldsFragment.js
thomasgwatson: It'll just return an empty set. Is there a bookshelf.js way to say "only look this up if the post is type === proposal"?

src/components/PostCard/PostBodyProposal/PostBodyProposal.js
thomasgwatson: done

src/graphql/fragments/postFieldsFragment.js
tibetsprague: yes, see my comment in hylo-node

thomasgwatson:
Screenshot 2024-05-09 at 6 06 26 PM
Screenshot 2024-05-09 at 6 06 51 PM
Screenshot 2024-05-09 at 6 07 12 PM
Screenshot 2024-05-09 at 6 07 29 PM

An interface for adding proposal outcomes. Visible only to authors.

The outcome display below the quorum bar is visible to all

tibetsprague:
that looks good to me and I am happy to go with it. I thought we had landed on the outcome text just being a pinned comment but once again our notes aren't clear and I am good with this version.

src/components/PostEditor/PostEditor.js
tibetsprague: i still disagree here about this, i guess we could bring to the group. also 100 is listed twice :) and does the order matter? if you are going to have double emojis then i think the single version should go right before the double one.

public/locales/es.json
```diff
@@ -865,6 +868,7 @@
   "Word used to describe a group Moderator": "Palabra utilizada para describir a un moderador de grupo",
   "Would you like to receive a notification for each new post in your groups?": "¿Te gustaría recibir una notificación por cada nueva publicación en tus grupos?",
   "Write something...": "Escribe algo...",
+  "Yes, I am ready to summarize": "Sí, estoy listo para resumir.",
```
tibetsprague, line 871: to discuss

tibetsprague:
i think this is at a point where we should start testing it on staging, that will be the best way to feel it out and see if we need to make more changes.

thomasgwatson:
Yeah for sure, lets get it up on staging

tibetsprague:
Trying to create a proposal post on staging errored out

Screenshot 2024-05-16 at 8 45 07 PM

thomasgwatson:
Can you check the network tab? Any warnings?

tibetsprague:
Just said an error in createPost. Staging doesnt show the errors unfortunately.

thomasgwatson:
Doing a proposal pretty much like your succeeds in local env but fails in staging. Not the clean debugging situation I would have liked. Can we get into the logs of the staging instance?

Screenshot 2024-05-16 at 11 14 19 PM
Screenshot 2024-05-16 at 11 14 29 PM

tibetsprague:
Do you have Heroku access? You should if you dont

thomasgwatson:
Ahhh I think I do. Sorry, I'll check it out

tibetsprague:
Also I can try to update the database from prod in the morning if we cant figure out whats up. But could be some production setting that is different from local. An environment variable?

thomasgwatson:
I don't think I've added or changed any env variables. I'll have a dig
later, heading out

thomasgwatson:
Well, the logs are unfortunately completely uninteresting.

No error message. GraphQL log seems to be fine...


Screenshot 2024-05-17 at 4 06 16 PM

Screenshot 2024-05-17 at 4 06 07 PM

tibetsprague:
Starting to put in feedback/requests from testing: #1) Quorum should show the number of people that the percentage represents. Though this could be tricky if multiple groups are selected... hmm what do we do with quorum if multiple groups are selected??

tibetsprague:
Seeing all kinds of weird behavior if i select a template then delete all the options and then select a different template. Actually selecting templates only seems to work sometimes, not sure why. Certain templates work more often than others. I think maybe we need to walk through all the templates as a group and decide what they do. For example I dont know about Poll changing the post title. Also consent doesnt necessarily mean quorom = 100%, it normally means no blocks among the quorum, which could even be 0. Just no one blocks over a certain period of time.

thomasgwatson:
Are there specific conditions you can recreate that trigger the weird
behavior with templates? I just whipped down the list of them, selecting
each and then deleting all the options and it worked as I expected it.

Certainly, until we add 'weights' to options (or otherwise make 'block' or
'veto' magic options), there are a few templates that don't truly represent
what they are called. All the quorums are suggestions, you can change them
all you like. Similarly, poll going into the title is just a suggestion.

Quorum for multiple groups does add up all the membership numbers for all
the groups.

const calcNumberOfPossibleVoters = (groups) => {
return groups.reduce((acc, group) => {
return acc + group.memberCount
}, 0)
}

I would say it's *rare* that a multiple group quorum would make sense but
it's possible that it could.

tibetsprague:
undoing proposal complete should remove the outcome (i think?) but probably have a warning alert first

tibetsprague:
if proposal is complete, votes should not be able to be changed

tibetsprague:
governance view showing all post types right now. also remove the post type dropdown

tibetsprague (approved):

PR #1656, opened 2024-07-25
Evo 1654 fix prev and next arrows on slick photo carousel

### LINKED ISSUE #1654: Add modal keyboard events handling
Problems:
1) The prev arrow is past the edge of the box and
2) the keyboard arrows don't work

KevinTriplett:
fixes #1654 

Three commits:
- fix position of arrows on image carousel
- add modal keyboard and mouse events
-- previously, only the background DOM received events
-- escape now cancels the modal
-- return now submits the modal (NOTE: if no submit action, just closes modal)
-- clicking anywhere outside the modal cancels it
- add image carousel keyboard events
-- previously, only the background DOM received events
-- after above commit, modal received events but not image carousel
-- NOTE: only way to trigger slick-slide without a ref is to trigger mouse events on the arrow buttons

Best I could do quickly -- may have been a cleaner way to do it but eh

### DIFF
--- a/src/components/ImageCarousel/ImageCarousel.js
+++ b/src/components/ImageCarousel/ImageCarousel.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useRef, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import Slider from 'react-slick'
 import { filter, isEmpty } from 'lodash/fp'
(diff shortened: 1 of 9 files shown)


KevinTriplett:
Just realized this branch is based off an issue branch and not off of dev

PR #1657, opened 2024-07-25
Evo 1654 add modal keyboard events handling

### LINKED ISSUE #1654: Add modal keyboard events handling

KevinTriplett:
fixes #1654 

Problems:
- "prev arrow" icon of image carousel was off screen / clipped
- image carousel did not respond to left arrow and right arrow keyboard action
- modal did not respond to escape and return keyboard action
- clicking outside the modal did not dismiss the modal, an expected action for most users

Three commits:
- fix position of "prev arrow" on image carousel
- enable modal keyboard and mouse events
  - previously, only the background DOM received events
  - escape now cancels the modal
  - return now submits the modal (NOTE: if no submit action, just closes modal)
  - clicking anywhere outside the modal cancels it
- enable image carousel keyboard events
  - previously, only the background DOM received events
  - after above modal commit, slick-slide in image carousel still did not receive keyboard events
  - NOTE: only way to trigger slick-slide without a ref is to trigger mouse events on the arrow buttons

Best I could do quickly -- may have been a cleaner way to do it but eh


src/components/ModalDialog/ModalDialog.js
```diff
@@ -61,13 +61,31 @@ class ModalDialog extends Component {
     useNotificationFormat: false
   }
 
+  modalRef = createRef()
+
+  handleKeydown = event => {
+    if (event.code === 'Escape') this.submit()
```
tibetsprague, line 67: uh shouldn't this be cancel?

src/components/ModalDialog/ModalDialog.js
```diff
@@ -61,13 +61,31 @@ class ModalDialog extends Component {
...
+    if (event.code === 'Enter') this.cancel()
```
tibetsprague, line 68: and this submit?

tibetsprague (changes requested):

src/components/ModalDialog/ModalDialog.js
KevinTriplett: Ooh thanks ... yes, I recall that I originally was using event.keyCode and then switches to event.code ... and didn't switch the calls. Commit pushed

tibetsprague (approved):

KevinTriplett:
I'm looking at tests I can add to improve coverage and catch future regressions

KevinTriplett:
@tibetsprague if you want to look over the recent commits, I simplified the keyboard event handling tremendously and had to update some other files to get eslint to pass.

Which I assume means maybe you and Thomas don't have eslint automatically check work before pushing?

PR #1660, opened 2024-07-26
Fix invalid post warning when changing post types

### LINKED ISSUE #1655: Post button inappropriately mentions proposal options
How to replicate:
1) Start creating an Event type post
2) Change post type to Discussion
3) Add title and text but Post button stays disabled and tooltip talks about needing proposal options
4) Chanage post type to Offer and back to Discussion and Post button is enabled.

KevinTriplett:
Fixes #1655 

Hi Thomas -- again, I'm impressed with the work that has gone into proposals.

The root for this issue seems to be that isProposal is updated when the component renders but not before? It seems like isProposal is correct when the component renders, but not when `isValid` gets called.

(Moving the invalidPostWarning variable initialization to the PostEditor component was my first try but had to leave it there, not sure why. Putting it back into the ActionBar component class caused the message to not update at all.)


thomasgwatson (approved):
lgtm...

KevinTriplett:
@thomasgwatson FYI discovered logic error and saw missing coverage for valid/invalid tests.

So I fixed the logic, protected against nulls, and wrote additional tests.

src/components/PostEditor/PostEditor.js
```diff
@@ -447,19 +447,26 @@ class PostEditor extends React.Component {
       this.state.post,
       postUpdates
     )
-    const { isEvent, isProject, isProposal } = this.props
+
+    let validEvent = type?.length > 0
```
tibetsprague, line 451: shouldn't this be called validPost?

KevinTriplett: Ah, yikes ... This deals with post type data only (there are other tests for valid post), so I renamed it validTypeData.

tibetsprague (approved):

PR #1666, opened 2024-07-31
Reload current user whenever looking at the group detail page, in cas…

### LINKED ISSUE #1663: Prompted to re-answer join questions once my request to join a group approved
Another thing, I requested access to join BioFi. I answered the join questions and requested access. Once my request was approved, I'm taken to the Explore page and immediately I'm prompted to re-answer the join questions. It won't let me hop in until I do so. 

This is happening on my Macbook Air, on Chrome

thomasgwatson:
…e membership has changed

closes #1663

### DIFF
--- a/src/routes/GroupDetail/GroupDetail.connector.js
+++ b/src/routes/GroupDetail/GroupDetail.connector.js
@@ -10,6 +10,7 @@ import getGroupForDetails from 'store/selectors/getGroupForDetails'
 import getResponsibilitiesForGroup from 'store/selectors/getResponsibilitiesForGroup'
 import { FETCH_GROUP_DETAILS } from 'store/constants'
 import getRouteParam from 'store/selectors/getRouteParam'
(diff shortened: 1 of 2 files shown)


tibetsprague (approved):
seems fine, and this fixes the double answering of questions? seems like that happens when going to the group home/stream page, not group details. but this happens first?

thomasgwatson:
We could add a similar refetch in the group welcome modal as well, as a

I tried accessing the group in a range of ways (clicking on the notif,
group explorer) and those didn't get to me to place where 1) I had a
membership 2) but the membership didn't have a `joinQuestionsAnsweredAt`
value

thomasgwatson:
(And when you don;t have a group membership, it routes you to GroupDetail

tibetsprague:
as long as it fixes the specific case Krisha ran into, which I have also seen then this is good enough for now

PR #1669, opened 2024-08-09
Evo 1667 image carousel not listening to keyboard events

### LINKED ISSUE #1667: Image carousel not listening to keyboard events and aspect ratio off
When an image is clicked from a PostCard, the keyboard events are not connecting to the listener, only after the PostDetails are showing on the right-hand side and the PostCard image is clicked are the listeners connected.

So it must be kind of like a race condition.

KevinTriplett:
Closed #1667 

Two commits:
- removed an onClick that overrode the CardImageAttachments onClick
  - this means that clicking an image on a PostCard opens the ImageCarousel modal and not the PostDetail slide-out (so this might not be the way we want it to behave, we may want it to open the PostDetails and require another click/touch to open the ImageCarousel modal)
- fix a WebKit bug that caused the "others" thumbnail images to stretch

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "dev": "yarn start",
     "start": "rm -rf ./node_modules/.cache/babel-loader && node --preserve-symlinks scripts/start.js",
     "build": "node --max-old-space-size=4000 scripts/build.js --stats && scripts/build-es5.sh",
-    "test": "node scripts/test.js",
+    "test": "TZ=UTC node scripts/test.js",
     "lint": "yarn exec standard",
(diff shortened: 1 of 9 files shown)


KevinTriplett:
Third commit to fix tests failing CI due to difference in timezone

KevinTriplett:
Okay, fixed now and ready for review

KevinTriplett:
I'm going to mothball this, I just realized that clicking the image in the Card view should show the post details, not the image carousel.

tibetsprague (commented):
ooh yay thank you!

PR #1675, opened 2024-08-11
Evo 1672 image carousel opening when post is clicked in card view

### LINKED ISSUE #1672: Image carousel opening when post is clicked in card view
When clicking on a post image in card view, and image carousel opens in addition to the post details. This is incorrect behavior, only the post details should open.

KevinTriplett:
Closes #1672 

- prevent ImageCarousel from opening when clicking an image attached to a PostCard
- correct a typo: change onKeydown to onKeyDown

Not sure why I hadn't noticed this typo before now, but including it here since it's tangentially involved.


KevinTriplett:
keep the class name case consistent

tibetsprague (approved):

